PR #960, opened 2022-02-02
ELN Publishing (Backend)

thomas-vu:


yattias (approved):
Looks good. Would be good to have some basic tests related to publishing. These can come after we merge in a subsequent PR.  Testing the `update_existing_researchhub_posts` endpoint by ensuring only the author can publish would be good.

PR #963, opened 2022-02-03
Prevent basic user from being able to delete hubs

yattias:
### What?

- Prevent basic users from being able to delete hubs
- Tests

### DIFF
--- a/src/hub/permissions.py
+++ b/src/hub/permissions.py
@@ -28,7 +28,7 @@ class CensorHub(RuleBasedPermission):
 
     def satisfies_rule(self, request):
         if request.method == "DELETE":
-            return request.user.is_anonymous is not True and request.user.is_authenticated and request.user.moderator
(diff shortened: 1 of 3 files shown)


src/hub/permissions.py
```diff
@@ -28,9 +28,9 @@ class CensorHub(RuleBasedPermission):
 
     def satisfies_rule(self, request):
         if request.method == "DELETE":
-            return request.user.is_authenticated and request.user.moderator
+            return request.user.is_authenticated and request.user.moderator and request.user.is_anonymous is not True
```
le-sun, line 31: It's best to check if the user is anonymous first because if the user is anonymous, the fields `is_authenticated`  and `moderator` won't exist and will result in an error.

le-sun (changes requested):

src/hub/permissions.py
yattias: Good point 👍

le-sun (approved):
Looks good

PR #964, opened 2022-02-03
Complete pending payouts

lightninglu10:
Some users were hit with a bug where their withdrawal did not fire. This is a management command to update those pending payouts.

### DIFF
--- a/src/reputation/management/commands/complete_pending_payout.py
+++ b/src/reputation/management/commands/complete_pending_payout.py
@@ -1,0 +1,26 @@
+from django.core.management.base import BaseCommand
+from purchase.models import Balance
+from reputation.lib import (
+    PendingWithdrawal
+)
+from reputation.models import Withdrawal
+from django.contrib.contenttypes.models import ContentType
+
+class Command(BaseCommand):
+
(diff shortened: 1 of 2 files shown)


src/reputation/management/commands/complete_pending_payout.py
```diff
@@ -0,0 +1,26 @@
...
+from reputation.lib import (
+    PendingWithdrawal
+)
+from reputation.models import Withdrawal
+from django.contrib.contenttypes.models import ContentType
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        withdrawals_to_complete = [1868, 1864]
```
yattias, line 12: It seems like these transactions are hard-coded which will give us a one-time usage out of this command. Does it make sense to accept the transactions as an arg instead?

lightninglu10: I think we should just change this everytime we need to run it. An arg would be OK too if we see us using this a lot.

yattias: Sounds good

yattias (approved):

PR #967, opened 2022-02-08
[Permissions] Allow Appropriate Editors to censor Unified Doc

calvinhlee23:
- Updated Endpoints
- Updated Permissions
- added integration tests

TEST: 
- run ```python manage.py test researchhub_document.tests.test_view```
Screen Shot 2022-02-09 at 2 57 51 PM


lightninglu10:
@calvinhlee23 is this ready to merge? has it been tested on staging?

src/researchhub_document/permissions.py
```diff
@@ -1,29 +1,52 @@
...
+        if (doc is None):
+            return False
 
-        doc = model.objects.get(unified_document_id=obj.id)
-        if doc.created_by_id == request.user.id or request.user.moderator:
+        requestor = request.user
+        is_requestor_appropriate_editor = is_hub_editor(
+            requestor,
+            doc.hubs,
+        )
```
le-sun, line 44: Let's make this a method on the user class, so we can just call `user.is_hub_editor()`

src/researchhub_document/permissions.py
```diff
@@ -1,29 +1,52 @@
...
 
 class HasDocumentCensorPermission(AuthorizationBasedPermission):
     message = 'Need to be author or moderator to delete'
 
     def is_authorized(self, request, view, obj):
         if request.user.is_authenticated is False:
             return False
+        
+        document_type = obj.document_type if obj.__class__ is \
+            ResearchhubUnifiedDocument else Paper
```
le-sun, line 27: Use `isinstance`

src/researchhub_document/permissions.py
```diff
@@ -1,29 +1,52 @@
...
+        doc = None
+        
+        if document_type == DISCUSSION:
+            doc = ResearchhubPost.objects.get(unified_document_id=obj.id)
+        elif document_type == HYPOTHESIS:
+            doc = Hypothesis.objects.get(unified_document_id=obj.id)
+        else:
+            doc = Paper.objects.get(id=obj.id)
```
le-sun, line 35: Instead of querying the database again for the object, we can just use the passed obj parameter.

Ex:
```
doc = obj
obj.get_document() # Create a class method to retrieve the document

# Class Method
if self.document_type == 'PAPER':
    doc = self.paper
elif ...
```

le-sun (changes requested):
@yattias This includes some suggestions I was going to make for your PR

src/researchhub_document/permissions.py
yattias: Maybe I'm missing something but I'm not a fan of defaulting to Paper if not unified doc instance. I would prefer if we do explicit checks to help mitigate bugs in the future.

yattias (commented):
@calvinhlee23 I made one comment.
Do you mind adding a couple of unit tests? My PR has a bunch for hypothesis and posts. Should be pretty quick to add a couple in if you follow the examples

src/researchhub_document/permissions.py
calvinhlee23: your suggestion was what was there before.
I made changes like this because when a request was made against a paper, it comes in as a paper instance not a unidoc which means the "object" doesn't have a field document_type, resulting in an error. 
If i have a try-catch (which i also tried, the code looks messy) & if i had some sort of a check to filter it out, the code becomes redundant.

lightninglu10 (approved):
@le-sun if this is good to go from your side we should get it merged & deployed

le-sun (approved):
Don't forget to remove dump.rdb

PR #976, opened 2022-02-15
Payout email fix

calvinhlee23:
### Code: 
- Changing the priority bucket seems to have fixed the double-email issue. 

### Test: 

Screen Shot 2022-02-14 at 6 14 29 PM
Screen Shot 2022-02-14 at 6 14 26 PM


le-sun (approved):
Interesting that changing the priority fixes this issue. Doesn't really make sense why but at least it works.

PR #982, opened 2022-02-22
Transactions return source as object

lightninglu10:
Returning nested source so that the frontend can render paper title and slug


src/reputation/serializers.py
```diff
@@ -25,7 +44,54 @@
...
+            ).data
+        elif isinstance(value, Purchase):
+            return PurchaseSerializer(value, context={'exclude_stats': True}).data
+
+        sentry.log_info('No representation for {} / id: {}'.format(str(value), value.id))
+        return None
+
+
 class DistributionSerializer(serializers.ModelSerializer):
+    proof_item = ProofRelatedField(read_only=True)
```
yattias, line 93: I see this not being used. Is that intentional?

yattias (approved):
LGTM, just one comment

src/reputation/serializers.py
lightninglu10: this is being used -- it's returned with the serializer

PR #983, opened 2022-02-22
Remove pending from balance

lightninglu10:
Remove pending amount from balance, do not allow a user to withdraw pending.


le-sun (approved):
Looks good

PR #985, opened 2022-02-25
Hot Score V2

yattias:
### What?

- Hot Score V2 algorithm
    - The basic idea is that the score of a document depends on the sum of various interactions on it (i.e. votes). Each of which, depends on **how recent** these interactions were. The more recent the interaction is, the greater its value. 
- The algorithm ships with a verbose endpoint `/api/researchhub_unified_documents/{doc_id}/hot_score/?debug` to help debug the hot score for a given document
- In addition, there is a script `recalc_hot_score` which will be used to recalculated the new hot score

### Important
- The old hotscore will not be affected in any way and will continue to be used on the FE until we are ready to switch over.
  - Technically speaking, there is a new field on unified document called `hot_score_v2` which will hold the new value 
- To see a preview of new feed results hotscore, pass `hot_v2` flag to the unified feed endpoint


src/researchhub_document/related_models/researchhub_unified_document_model.py
```diff
@@ -87,6 +92,18 @@ def authors(self):
...
 
+    def get_document(self):
+        if (self.document_type.upper() == 'PAPER'):
+            return self.paper
+        elif (self.document_type.upper() == 'DISCUSSION'):
+            return self.posts.first()
+        elif (self.document_type.upper() == 'HYPOTHESIS'):
+            return self.hypothesis
+        elif (self.document_type.upper() == 'NOTE'):
+            return self.note
```
le-sun, line 103:
```
from researchhub_document.related_models.constants.document_type import (
  Discussion,
  Note,
  ...
)
```

```suggestion
        if self.document_type == PAPER):
            return self.paper
        elif self.document_type == DISCUSSION):
            return self.posts.first()
        elif self.document_type == HYPOTHESIS):
            return self.hypothesis
        elif self.document_type == NOTE:
            return self.note
```

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -100,6 +100,29 @@ def restore(self, request, pk=None):
             status=200
         )
 
+    @action(
+        detail=True,
+        methods=['get'],
+        permission_classes=[AllowAny],
+    )
+    def hot_score(self, request, pk=None):
+        debug = True if request.query_params.get('debug') is not None else False
```
le-sun, line 109:
```suggestion
        debug = request.query_params.get('debug', False)
```

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -117,6 +140,7 @@ def update(self, request, *args, **kwargs):
         return update_response
 
     def _get_document_filtering(self, query_params):
+        use_v2_hot_score = query_params.get('hot_v2') == 'true'
```
le-sun, line 143:
```suggestion
        use_v2_hot_score = query_params.get('hot_v2', False)
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+            print(f'Value: {final_val}')
+
+        return final_val
+
+    def _calc_social_media_score(self):
+        social_media_score = 0
+        doc = self.get_document()
+
+        date_val = self._get_date_val(self.created_date)
+        if self.document_type.upper() == 'PAPER':
```
le-sun, line 48:
```
from researchhub_document.related_models.constants.document_type import PAPER
```
```suggestion
        if self.document_type == PAPER:
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+            twitter_score = math.log(doc.twitter_score+1, 10)
+            social_media_score = date_val * twitter_score
+
+        return social_media_score
+
+    def _calc_vote_score(self, votes):
+        return sum([
+            self._get_date_val(v.created_date) for v in votes
+        ])
```
le-sun, line 57:
```suggestion
        return sum(
            map(
                self._get_date_val,
                votes.values_list('created_date', flat=True)
            )
        )
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+        debug_obj = {
+            'unified_doc_id': self.id,
+            'inner_doc_id': doc.id,
+            'document_type': self.document_type,
+            'DISCUSSION_VOTE_WEIGHT': DISCUSSION_VOTE_WEIGHT,
+            'DOCUMENT_VOTE_WEIGHT': DOCUMENT_VOTE_WEIGHT,
+        }
+
+        # Doc vote score
+        if self.document_type.upper() == 'PAPER':
```
le-sun, line 81:
```suggestion
        if self.document_type == PAPER:
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+        debug_obj['doc_created_score'] = {'created_date': self.created_date, '=doc_created_score': doc_created_score}
+
+        # Doc social media score
+        social_media_score = self._calc_social_media_score()
+        debug_obj['social_media_score'] = {'=social_media_score': social_media_score}
+
+        # Doc discussion vote score
+        discussion_vote_score = 0
+        debug_obj['discussion_vote_score'] = {}
+        for t in doc.threads.filter(is_removed=False):
```
le-sun, line 101:
```suggestion
        for t in doc.threads.filter(is_removed=False).iterator():
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+            thread_vote_net_score = max(0, t.calculate_score())
+            thread_vote_time_score = self._calc_vote_score(t.votes.all())
+            thread_vote_score = thread_vote_net_score * thread_vote_time_score * DISCUSSION_VOTE_WEIGHT
+            discussion_vote_score += thread_vote_score
+
+            debug_val = {'vote_net_score': thread_vote_net_score, 'vote_time_score': thread_vote_time_score, '=thread_vote_score': thread_vote_score}
+            debug_obj['discussion_vote_score'][f'thread (id:{t.id})'] = debug_val
+            for c in t.comments.filter(is_removed=False):
```
le-sun, line 109:
```suggestion
            for c in t.comments.filter(is_removed=False).iterator():
```

src/researchhub_document/hot_score_mixin.py
```diff
@@ -0,0 +1,142 @@
...
+                comment_vote_net_score = max(0, c.calculate_score())
+                comment_vote_time_score = self._calc_vote_score(c.votes.all())
+                comment_vote_score = comment_vote_net_score * comment_vote_time_score * DISCUSSION_VOTE_WEIGHT
+                discussion_vote_score += comment_vote_score
+
+                debug_val = {'vote_net_score': comment_vote_net_score, 'vote_time_score': comment_vote_time_score, '=comment_vote_score': comment_vote_score}
+                debug_obj['discussion_vote_score'][f'thread (id:{t.id})'][f'comment (id:{c.id})'] = debug_val
+                for r in c.replies.filter(is_removed=False):
```
le-sun, line 117:
```suggestion
                for r in c.replies.filter(is_removed=False).iterator():
```

le-sun (changes requested):

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: @le-sun I need here explicit check of 'true' string. I want to avoid the FE potentially adding the query var with no val yet still resulting in v2 feed results

src/researchhub_document/hot_score_mixin.py
lightninglu10: what's the point of calling .iterator here? is it to not load everything into memory?

PR #987, opened 2022-02-28
Query Optimize: Most Discussed

yattias:
### What?
- Removing queryset null check improves most-discussed query performance by 100X


le-sun (approved):
Looks good

PR #990, opened 2022-03-03
RSC Withdraw Rewrite

lightninglu10:


src/user/serializers.py
```diff
@@ -700,7 +700,12 @@ def get_item(self, action):
         if type(item) in ignored_items:
             return None
 
-        if isinstance(item, Paper):
+        if isinstance(item, Withdrawal):
+            from reputation.serializers import WithdrawalSerializer
+            serializer = WithdrawalSerializer
+            context = {}
+            _context_fields = {}
```
le-sun, line 707: Let's make this into a dynamic serializer to keep things consistent.

src/ethereum/lib.py
```diff
@@ -1,6 +1,8 @@
 from decimal import Decimal
 from ethereum.utils import decimal_to_token_amount
-from researchhub.settings import WEB3_RSC_ADDRESS
+from researchhub.settings import WEB3_RSC_ADDRESS, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, WEB3_KEYSTORE_BUCKET, WEB3_KEYSTORE_FILE, WEB3_KEYSTORE_PASSWORD, w3
```
le-sun, line 3:
```suggestion
from researchhub.settings import (
  WEB3_RSC_ADDRESS,
  AWS_ACCESS_KEY_ID,
  AWS_SECRET_ACCESS_KEY,
  WEB3_KEYSTORE_BUCKET,
  WEB3_KEYSTORE_FILE,
  WEB3_KEYSTORE_PASSWORD,
  w3
)
```

src/reputation/lib.py
```diff
@@ -31,28 +634,25 @@ def _calculate_tokens_and_update_withdrawal_amount(self):
...
-            "public_key": public_key
-        }
-        response = http_request(
-            RequestMethods.POST,
-            url,
-            data=json.dumps(data),
-            timeout=3
+
+        contract = w3.eth.contract(abi=contract_abi, address=Web3.toChecksumAddress(RSC_CONTRACT_ADDRESS))
+        amount = int(self.amount)
```
le-sun, line 646: Should this be a decimal?

src/reputation/models.py
```diff
@@ -124,6 +124,12 @@ def set_withdrawal(self, withdrawal_instance):
         self.withdrawal = withdrawal_instance
         self.save()
 
+class Webhook(models.Model):
+    body = JSONField(blank=True)
```
le-sun, line 128: Missing `null=True`?

src/reputation/models.py
```diff
@@ -124,6 +124,12 @@ def set_withdrawal(self, withdrawal_instance):
...
+    created_date = models.DateTimeField(auto_now_add=True)
+    updated_date = models.DateTimeField(auto_now=True)
+    from_host = models.CharField(max_length=64, blank=True)
```
le-sun, line 131: Missing `null=True`?

src/reputation/views.py
```diff
@@ -136,6 +139,53 @@ def get_queryset(self):
             return Withdrawal.objects.all()
         else:
             return Withdrawal.objects.filter(user=user)
+    
+    @action(
+        detail=False,
+        methods=['POST'],
+        permission_classes=[],
```
le-sun, line 146: Is this endpoint safe to be open to the public or do we want some extra permissions on it?

le-sun (changes requested):

src/user/serializers.py
lightninglu10: I'm not sure how to do that

src/reputation/lib.py
lightninglu10: we're only allowing users to put withdraw integer amounts at the moment

src/reputation/models.py
lightninglu10: The webhook shouldn't be null, if it is then it should error

src/reputation/models.py
lightninglu10: from_host also shouldn't be null

src/user/serializers.py
le-sun: In `reputation/serializers.py` add
```
class DynamicWithdrawalSerializer(DynamicModelFieldSerializer):
    user = serializers.SerializerMethodField()

    class Meta:
        model = Withdrawal
        fields = '__all__'

    def get_user(self, withdrawal):
        context = self.context
        _context_fields = context.get('rep_dws_get_user', {})
        serializer = DynamicUserSerializer(
            withdrawal.user,
            context=context,
            **_context_fields
        )
        return serializer.data
```

then

```suggestion
            from reputation.serializers import DynamicWithdrawalSerializer
            serializer = DynamicWithdrawalSerializer
```

lightninglu10: I'll update this  & test it in a tech debt sprint

PR #993, opened 2022-03-04
Cache invalidation fix for most discussed papers

yattias:
### What?
- Adding a fix to invalidate cache based on the right cache key


le-sun (approved):
Looks good

PR #995, opened 2022-03-09
Async Paper Flow

le-sun:
Package Changes:
- Added [Black](https://github.com/psf/black) (for formatting)
- Added [isort](https://github.com/PyCQA/isort) (for formatting)
- Updated Manubot
- Updated Habanero (crossref)

Updated pre-commit to auto format
Added Paper submission model for asynchronous flow
Added celery tasks to handle paper submission process
Updated notifications to handle paper submissions


src/paper/manubot.py
```diff
@@ -0,0 +1,59 @@
...
+
+CSLItem = Dict[str, Any]
+
+url_retrievers = [
+    get_url_csl_item_zotero,
+    get_url_csl_item_manual,
+]
+
+
+class RHCiteKey(CiteKey):
```
le-sun, line 17: Main

src/paper/models.py
```diff
@@ -1256,48 +1029,124 @@ class Flag(models.Model):
...
-        'user.User',
-        on_delete=models.CASCADE,
-        related_name='featured_papers'
+        "user.User", on_delete=models.CASCADE, related_name="featured_papers"
     )
     ordinal = models.IntegerField(default=0)
     created_date = models.DateTimeField(auto_now_add=True)
+
+
+class PaperSubmission(DefaultModel):
```
le-sun, line 1072: Main

src/paper/serializers.py
```diff
@@ -1044,64 +952,98 @@ class BookmarkSerializer(serializers.Serializer):
...
 
 
 class DynamicFigureSerializer(DynamicModelFieldSerializer):
     class Meta:
-        fields = '__all__'
+        fields = "__all__"
         model = Figure
+
+
+class PaperSubmissionSerializer(serializers.ModelSerializer):
```
le-sun, line 1014: Main

src/paper/tasks.py
```diff
@@ -1233,11 +1173,222 @@ def pull_crossref_papers(start=0, force=False):
...
     info = f"""
         Crossref Duplicates Detected: {num_duplicates}\n
         Total Crossref pull: {total_results}
     """
     sentry.log_info(info)
     return total_results
+
+
+# @app.task(bind=True, queue=f"{APP_ENV}_cermine_queue")
+@app.task(bind=True)
```
le-sun, line 1187: Main

src/paper/views.py
```diff
@@ -1362,18 +1183,43 @@ def get_vote_response(vote, status_code):
...
-    vote = Vote.objects.create(
-        created_by=user,
-        paper=paper,
-        vote_type=vote_type
-    )
+    vote = Vote.objects.create(created_by=user, paper=paper, vote_type=vote_type)
     return vote
+
+
+class PaperSubmissionViewSet(viewsets.ModelViewSet):
```
le-sun, line 1196: Main

le-sun (commented):
Comments with the word `Main` are sections of the code with the main async update. Everything else is minor/formatting.

yattias:
@le-sun now that I pushed the queue changes. do you mind specifying a queue for your tasks? If not specified, they will be mapped to `default` queue but we should specify a queue to avoid that when possible

le-sun:
> @le-sun now that I pushed the queue changes. do you mind specifying a queue for your tasks? If not specified, they will be mapped to `default` queue but we should specify a queue to avoid that when possible

Yep, I updated it to use a new queue.

calvinhlee23 (changes requested):
Can you try rebasing against current master? I feel like a lot of these infra code should be resolved when you rebase

le-sun:
> Can you try rebasing against current master? I feel like a lot of these infra code should be resolved when you rebase

I've added some packages that created these infra changes to hopefully create some more consistency within the codebase.

src/paper/serializers.py
```diff
@@ -577,41 +535,30 @@ def update(self, instance, validated_data):
...
-                    map(lambda hub: hub.id, remove_hubs + new_hubs)
-                )
+                updated_hub_ids = list(map(lambda hub: hub.id, remove_hubs + new_hubs))
                 if len(updated_hub_ids) > 0:
                     reset_unified_document_cache(
                         hub_ids=updated_hub_ids,
-                        document_type=['paper', 'all'],
+                        document_type=["paper", "all"],
                         filters=[NEWEST, TOP, TRENDING, DISCUSSED],
+                        with_default_hub=True,
```
yattias, line 544: I recommend you do not include this here in this case

le-sun: When the async flow finishes and the user has to input a hub, the paper won't show up in the feed without a feed cache reset. Is there something else that I should do instead?

yattias: The reason behind this is that the homepage doesn't need to be updated that often. Given that it updates automatically every 30 mins does the trick in most cases..

There are times thought such as when a new paper is added though that we should update the HP feed cache immediately

PR #998, opened 2022-03-10
Allow Mods to send RSC

lightninglu10:
Endpoint to allow mods to send RSC


src/start-es.sh
```diff
@@ -6,4 +6,4 @@ docker run \
   --publish 9200:9200 \
   --publish 9300:9300 \
   --env "discovery.type=single-node" \
-  docker.elastic.co/elasticsearch/elasticsearch:7.4.1
+  docker.elastic.co/elasticsearch/elasticsearch:7.17.1
```
lightninglu10, line 9: Changed this version because 7.4.1 no longer worked on m1 macs

yattias (commented):
This looks good. Wondering if you can create a couple of tests to help ensure functionality doesn't break in the future. Something simple like testing that the
- Recipient's balance should reflect RSC sent
- Sender's balance should reflect RSC sent
- Basic user should not be able to send RSC

lightninglu10:
> This looks good. Wondering if you can create a couple of tests to help ensure functionality doesn't break in the future. Something simple like testing that the
> 
> * Recipient's balance should reflect RSC sent
> * Sender's balance should reflect RSC sent
> * Basic user should not be able to send RSC

Added tests

src/purchase/views.py
```diff
@@ -73,6 +75,41 @@ class BalanceViewSet(viewsets.ReadOnlyModelViewSet):
...
+                proof=proof,
+                recipient_id=recipient_id,
+                distributed_status=Distribution.DISTRIBUTED
+            )
+            purchase = Balance.objects.create(
+                user_id=recipient_id,
+                amount=amount,
+                content_type=content_type,
+                object_id=distribution.id
+            )
```
le-sun, line 110: We should use the Distributor class for this. See `reputation/views.py` line 73.

le-sun (changes requested):

src/purchase/views.py
lightninglu10: I looked at this and the Distributor is a bit complex for what we're doing here -- i think what we have here is good

le-sun: I disagree. What makes the distributor class complex for what this accomplishes? It does the exact same thing.

PR #1001, opened 2022-03-15
Hot Score Algo Update

yattias:
### What?
- Second attempt at hot score algo. WIP


src/researchhub_document/hot_score_mixin2.py
```diff
@@ -0,0 +1,151 @@
...
+    def _c(self, num):
+        if num >= 0:
+            return 1
+        else:
+            return -1
+
+    def _count_doc_comment_votes(self, doc):
+        vote_total = 0
+        try:
+            for t in doc.threads.filter(is_removed=False).iterator():
```
lightninglu10, line 23: do we have unified_doc linked to comments / threads / replies? if so then it would be easier to just filter by the unified doc or paper

lightninglu10 (approved):

src/researchhub_document/hot_score_mixin2.py
yattias: @lightninglu10 I don't think so but I shall check!

PR #1005, opened 2022-03-17
Turning on hot score v2 by default

yattias:
### What?
- Turning on hot score v2 by default
- Another PR will follow to turn off old call to `calculate_hot_score` (v1) once things look good in prod for a couple of weeks


le-sun (approved):
Looks good. Can we remove all instances of the previous calculate_hot_score mechanisms? Like `paper/models.py` `def calculate_hot_score`

yattias:
@le-sun I will issue another PR to do that after things look good for a bit in case we want to revert for some reason

PR #1008, opened 2022-03-18
Added Crossref DOIs to publishing (Backend)

thomas-vu:


src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -114,6 +114,13 @@ class ResearchhubPost(AbstractGenericReactionModel):
         related_query_name='posts'
     )
     slug = models.SlugField(max_length=1024)
+    doi = models.CharField(
+        max_length=255,
```
le-sun, line 118: Will DOIs ever get that long?

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -165,6 +207,10 @@ def update_existing_researchhub_posts(self, request):
             document_types=['all', 'posts']
         )
 
+        if assign_doi:
+            charge_doi_fee(created_by, rh_post)
```
yattias, line 211: I think we should flip the operation and charge only if DOI was assigned.

Also, are we handling error on the FE in case crossref fails?

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -185,3 +231,51 @@ def create_unified_doc(self, request):
...
+    files = {
+        'operation': (None, 'doMDUpload'),
+        'login_id': (None, CROSSREF_LOGIN_ID),
+        'login_passwd': (None, CROSSREF_LOGIN_PASSWORD),
+        'fname': ('crossref.xml', crossref_xml),
+    }
+    crossref_response = requests.post(CROSSREF_API_URL, files=files)
+
+
+def charge_doi_fee(created_by, rh_post):
```
le-sun, line 265: We need a check to ensure the user has enough RSC to assign a DOI.

le-sun (changes requested):

le-sun:
Also the automated tests are failing

src/researchhub_document/views/researchhub_post_views.py
thomas-vu: There is a check for it before calling this function.

yattias (changes requested):
Just one comment. Also, can we add a test? I understand that a test for adding a DOI is a bit tricky but perhaps one for **NOT** adding a DOI if the user doesn't check the box

yattias:
Also @thomas-vu it appears tests are failing here

src/researchhub_document/related_models/researchhub_post_model.py
thomas-vu: They shouldn't since we control how long the suffix is, but this matches the paper model's DOI field.

src/researchhub_document/views/researchhub_post_views.py
thomas-vu: Good point, I'll add the check and error handling.

thomas-vu: I added some test as well.

thomas-vu: I added some tests as well.

le-sun (approved):

PR #1013, opened 2022-03-29
Cache Optimization

yattias:
## Analysis of our caching issues:
Periodically, we “preload” various hub feed pages into cache so the user does not experience slowdown when visiting.
- These “preload” tasks have a priority of 1 which is the highest.
- Given that we have a lot of hubs, this process can take a while - 10mins+
- While this is ongoing, celery will ignore any other high/low priority jobs that are added to the queue
- For example, users visiting our homepage trigger a job to preload it into cache but it will be ignored until all the other hubs finished
- Until then, everyone visiting the HP will experience a `False` `cache_hit` 

### The solution here is two fold:
1. Change the priorities so that homepage always take precedent
2. Avoid preloading too often and also when preloading, do so first for hubs that have the most activity

### Things done
- Do not clear caches via `cache.delete` instead, just override cache keys. 
- Rather than clearing all caches on each upvote/downvote, only clear trending and top caches
  - To get the latest score, we merge fresh data with cache data
- Avoid clearing homepage caches on every qualified action. Instead, only do so when absolutely necessary and rely on a periodic task to rebuild the cache


src/researchhub_document/tasks.py
```diff
@@ -167,124 +170,37 @@ def preload_trending_documents(
...
+        document_type=["all"]
+    )
 
 
 @app.task
 def update_elastic_registry(post):
     registry.update(post)
-
-
-def invalidate_feed_cache(
```
yattias, line 240: We are deprecating this function in favor of `reset_unified_document_cache`

src/researchhub_document/utils.py
```diff
@@ -99,24 +99,15 @@ def reset_unified_document_cache(
         TOP
     ],
     date_ranges=CACHE_DATE_RANGES,
-    use_celery=True
+    with_default=False,
```
yattias, line 102: Do not clear homepage caches by default, only when explicitly specified

src/researchhub_document/utils.py
```diff
@@ -130,24 +121,9 @@ def reset_unified_document_cache(
                                 f,
                                 time_scope,
                             ),
-                            priority=1,
-                            countdown=1
+                            priority=2,
```
yattias, line 124: I lowered the priority a bit to (hopefully) avoid creating a bottleneck. @le-sun any thoughts on that?
Also, curious to know where `priority` is defined in the codebase since it seems to not be a native celery config

src/researchhub_document/tasks.py
```diff
@@ -167,124 +170,37 @@ def preload_trending_documents(
     return paginated_response.data
 
 
-# Executes every 5 minutes
 @periodic_task(
-    run_every=crontab(minute='*/25'),
+    run_every=crontab(minute='*/1'),
```
yattias, line 174: This is for testing purposes

src/paper/tasks.py
```diff
@@ -657,40 +657,6 @@ def celery_update_hot_scores():
         paper.calculate_hot_score()
 
 
-# Executes every 5 minutes
-@periodic_task(
-    run_every=crontab(minute='*/5'),
-    priority=1,
-    options={'queue': f'{APP_ENV}_core_queue'}
-)
-def celery_preload_hub_papers(hub_ids=None):
```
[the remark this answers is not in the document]
lightninglu10: this is running periodically every 5 minutes. the decorator @periodic_task specifies that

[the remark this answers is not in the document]
yattias: @lightninglu10 sorry, I meant that despite this running periodically, I don't think it does anything productive. Wanted to get confirmation in case I am missing something here

src/paper/tasks.py
```diff
@@ -1063,11 +1029,11 @@ def pull_papers(start=0, force=False):
 NUM_DUP_STOP = 30
 
 # Pull Daily
-@periodic_task(
-    run_every=crontab(minute=0, hour='*/6'),
-    priority=1,
-    options={'queue': f'{APP_ENV}_autopull_queue'}
-)
+# @periodic_task(
```
yattias, line 1032: This function returns immediately. I'm commenting out the periodic part to avoid celery picking it up

le-sun (changes requested):

src/researchhub_document/tasks.py
```diff
@@ -143,10 +142,14 @@ def preload_trending_documents(
         page,
         _include_fields=[
             'created_by',
+            'created_date',
             'documents',
             'document_type',
             'hot_score',
-            'score'
+            'hot_score_v2',
+            'discussed',
```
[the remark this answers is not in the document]
yattias: Good observation, this was a test. I forgot to remove this field

le-sun (approved):

PR #1033, opened 2022-04-06
Dynamic upvote RSC

lightninglu10:
Sets RSC upvote reward dynamically


src/paper/models.py
```diff
@@ -480,6 +480,16 @@ def votes_indexing(self):
             return [self.get_vote_for_index(vote) for vote in all_votes]
         return {}
 
+    def true_author_count(self):
+        registered_author_count = self.authors.count()
+        raw_author_count = len(self.raw_authors)
+
+        for author in self.raw_authors:
```
yattias, line 487: I'm not super sure but I think `self.raw_authors` can cause an exception. It appears that we allow it to be NULL in the db. So if this value does not exist, this statement would throw a `'NoneType' is not iterable`

src/paper/models.py
```diff
@@ -480,6 +480,16 @@ def votes_indexing(self):
...
+            if self.authors.filter(first_name=author.get('first_name'), last_name=author.get('last_name')).exists():
```
yattias, line 488: When I was working on elastic and had to run extract authors from this field I realized that it is not guaranteed that each author will be a `dict`. I ended up writing a function `parse_author_name` for this purpose. 

For your purpose though, you may want to do a check and make sure that the author is a `dict` similarly to what I do in `raw_authors_indexing` because sometimes it can be a string.

src/reputation/distributions.py
```diff
@@ -16,6 +23,90 @@ def amount(self):
     def gives_rep(self):
         return self._give_rep
 
+RSC_YEARLY_GIVEAWAY = 50000000
+MINUTES_IN_YEAR = 525960
```
yattias, line 27: Is this a mistake? there are 525,600 minutes in one year

lightninglu10: Yes that is a mistake

lightninglu10: Well, google tells me there are 525960 minutes in a year.

lightninglu10: Screen Shot 2022-04-07 at 9 43 53 AM

src/paper/models.py
lightninglu10: @le-sun I want to get this cleaned before it's released then. 

Our raw authors should be in the same format regardless -- it should be either Null, an empty array, or an array of dicts

src/reputation/distributions.py
yattias: lol interesting. It's pretty minor. I just multiplied 60*24*365. Maybe it is the average between regular year and leap year or something like that.

src/reputation/distributions.py
```diff
@@ -16,6 +23,90 @@ def amount(self):
     def gives_rep(self):
         return self._give_rep
 
+RSC_YEARLY_GIVEAWAY = 50000000
```
yattias, line 26: I take it this will probably change year to year?

lightninglu10: yes, this amount will be variable. I think it's OK to set at the code level for now. We could also put it into an env var.

src/reputation/distributions.py
```diff
@@ -16,6 +23,90 @@ def amount(self):
...
+MINUTES_IN_YEAR = 525960
+HOURS_IN_YEAR = MINUTES_IN_YEAR / 60
+DAYS_IN_YEAR = 365
+MONTHS_IN_YEAR = 12
+GROWTH = .2
+
+def calculate_upvote_rsc():
```
yattias, line 33: Just a suggestion, had to dig into one this method does. I think it calculates the RSC to distribute per one vote?
If that's the case, perhaps we can rename to something like `calculate_rsc_per_upvote` or `calculate_rsc_to_distribute_per_upvote` or `calc_rsc_per_upvote` or something like that

src/reputation/distributions.py
```diff
@@ -16,6 +23,90 @@ def amount(self):
...
+
+    if paper:
+        from reputation.distributor import Distributor
+        from researchhub_case.models import AuthorClaimCase
+        author_distribution_amount = distribution_amount * .75
+        distribution_amount *= .25 # authors get 75% of the upvote score
+        distributed_amount = 0
+        author_count = paper.true_author_count()
+
+        for author in paper.authors.all():
```
yattias, line 86: Would authors that have not claimed yet only start receiving RSC for upvotes after they claim? or would there be a pot on the side for them to tap into and receive it once claimed?

src/reputation/distributions.py
yattias: Nah i think it is good here for us to see

yattias:
@lightninglu10 code looks good - could we write some unit tests to make sure the upvote calculation works properly?

lightninglu10:
> @lightninglu10 code looks good - could we write some unit tests to make sure the upvote calculation works properly?

Yep

lightninglu10:
> @lightninglu10 does this address the FE issue we saw where people can now go in and claim their profile by changing the authors field?

Yes it does -- it requires an approved claim case

src/reputation/distributions.py
lightninglu10: No, we're setting a pot aside for the authors. Upon claiming, they get a share of that pot.

yattias:
@lightninglu10 I think at least some of the following tests are worth writing:
1. `test_upvoting_paper_distributes_rsc`
- The validates that when an upvote happens, author(s) / submitters get appropriate RSC
2. `test_upvoting_hypothesis_distributes_rsc` 
3. `test_upvoting_post_distributes_rsc`  
4. `test_upvoting_thread_distributes_rsc`  
5. `test_upvoting_comment_distributes_rsc`  
6. `test_upvoting_reply_distributes_rsc`  

And I think at least something like
7. `test_cannot_redistribute_rsc_for` 
- test that user cannot redistribute by upvoting => then downvoting => upvoting again

le-sun (changes requested):
Can you rebase off master? The requirement changes I've made for the async flow should fix the requirements issue when deploying.

yattias:
Looks good @lightninglu10 - is there a test we can add to avoid the double spend issue? where you upvote => downvote => upvote again - basically asserting that the second time you upvote, the user does not get extra RSC?

le-sun (changes requested):

PR #1038, opened 2022-04-07
Fixed paper serializer for search

thomas-vu:
Screen Shot 2022-04-07 at 12 41 19 PM

### DIFF
--- a/src/search/serializers/paper.py
+++ b/src/search/serializers/paper.py
@@ -49,7 +49,7 @@ class PaperDocumentSerializer(DocumentSerializer):
             paper = Paper.objects.get(id=hit['id'])
             slug = paper.slug
         except:
-            pass
+            log_error(e, 'A paper must have slug')
 
         return slug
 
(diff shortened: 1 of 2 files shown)


yattias (approved):
Looks good.

@le-sun should we use `UserSerializer` or `DynamicUserSerializer` here? I guess it depends on whether the frontend needs all of this data?

src/search/serializers/paper.py
```diff
@@ -42,8 +46,8 @@ def get_highlight(self, obj):
     def get_slug(self, hit):
         slug = ''
         try:
-            obj = Paper.objects.get(id=hit['id'])
-            slug = obj.slug
+            paper = Paper.objects.get(id=hit['id'])
+            slug = paper.slug
         except:
             pass
```
lightninglu10, line 52: can we put a sentry exception here? all errors will be swallowed

src/search/serializers/paper.py
```diff
@@ -56,6 +60,23 @@ def get_unified_doc_id(self, paper):
...
+            return UserSerializer(user, read_only=True).data
+        except Exception as e:
+            print(e)
+
+    def get_uploaded_date(self, hit):
+        try:
+            paper = Paper.objects.get(id=hit['id'])
+            return paper.uploaded_date
+        except Exception as e:
+            print(e)
```
lightninglu10, line 78: instead of printing these exceptions, can we log them? otherwise it'll be hard to find if something errors

PR #1055, opened 2022-04-14
[AuthorClaim] Case Pagination

calvinhlee23:
- see title


src/researchhub_case/views/author_claim_case_view.py
```diff
@@ -122,12 +124,24 @@ def _get_author_claim_cases_for_mods(self, request):
...
+                data={
+                    'count': paginator.count,
+                    'has_more': pagination.has_next(),
+                    'page': page_num,
+                    'result': AuthorClaimCaseSerializer(
+                        pagination.object_list,
+                        many=True
+                    ).data,
+                },
+                status=200
```
le-sun, line 142: Was there a reason to not use the default pagination method?

Ex:
```
            page = self.paginate_queryset(all_documents)
            serializer = self.serializer_class(page)
            serializer_data = serializer.data
            return self.get_paginated_response(serializer_data)
```

le-sun (changes requested):
Have you ran `pre-commit install`?

src/researchhub_case/views/author_claim_case_view.py
calvinhlee23: ran

PR #1063, opened 2022-04-19
#1062 CRUD for Community Reviews

### LINKED ISSUE #1062: CRUD for community review
- Create a `review` model which has `score`, `thread_id`
- Update `Thread`'s create endpoint to conditionally create review or comment based on `comment_type` arg
- Update `Thread` serializer to return a review

yattias:
Closes #1062


src/discussion/views.py
```diff
@@ -111,14 +120,36 @@ def create(self, request, *args, **kwargs):
                 BaseComment.CREATED_LOCATION_PROGRESS
             )
 
-        response = super().create(request, *args, **kwargs)
-        response = self.get_self_upvote_response(request, response, Thread)
-        hubs = list(unified_document.hubs.all().values_list('id', flat=True))
-        discussion_id = response.data['id']
+        thread_response = None
+        try:
```
yattias, line 124: This may not be the prettiest. 
I want to rollback and fail in case any of these things fail.
What this does is:
1. Create thread
2. Create Review
3. Sets review property of thread

lightninglu10: instead of overloading this viewset, i wonder if we should make two separate calls from the frontend

First call: create a review with a score
Second call: create a discussion with the review attached

lightninglu10: @le-sun not sure what you think here ^^

lightninglu10 (changes requested):
Everything looks good, I just wanted to open discussion up on the overloaded viewset

src/discussion/views.py
le-sun: Yeah, it's probably best to create a new endpoint for adding reviews.

yattias: @le-sun / @lightninglu10 
Yeah, I understand the concern but wanted to discuss a bit more to make sure we do the right thing

My main issue is that having two endpoints will complicate things on the FE and BE
i.e. what if the BE fails to create a comment? What if the BE fails to create a review? What if the FE creates a comment but then fails to create a review? We will then need to handle each of these states or put a blind eye to it.

But then, I do think there is a good point here where in the future we could have a review that is decoupled from a comment - similarly to how Amazon lets you leave a review without a comment. This is especially likely if we will have a quantitative review form that collects many different data points.

Suggestion: What if we were to create an endpoint called `review` that basically just creates a review and will also create an optional comment if one is supplied. That endpoint could live in its own viewset or within the unified doc viewset (I think one of these two make the most sense to me)

That way:
1. We simplify the frontend by not calling two endpoints
2. We leave the comment endpoint as is and not overload it
3. Depending on my question below, we create a relatively thin endpoint

Question:
- Is it possible (and also is it good practice) to have an endpoint call another endpoint? In this case, the review endpoint could call the thread's viewset create endpoint (since it has all the business logic). I read that best practice is to have a serializer do this sort of thing but not sure if this proposed approach is frowned upon.

lightninglu10: I like the approach of creating a model viewset for reviews, but then creating a new endpoint inside of that review model viewset that deals with reviews for comments. 

I don't like putting it in the serializer because I think our logic for creating a review model should not depend on creating comments. Once we overload these creation points it's really hard to debug sometimes if things go weirdly / wrong.

If we limit it to a custom endpoint then I think we're guarded here if we ever just want to create a basic CRUD review

lightninglu10: I haven't ever seen an endpoint call another endpoint. I think if we were doing it like that we should make custom methods and call those custom methods directly instead of passing request variables around everywhere

src/review/views/review_view.py
```diff
@@ -0,0 +1,130 @@
...
+                request,
+                response.data['thread'],
+                Thread,
+                is_thread=True
+            )
+        
+        return response
+
+
+    def update(self, request, *args, **kwargs):
```
yattias, line 107: WIP

src/discussion/views.py
yattias: @lightninglu10 / @le-sun I spent more time thinking about this. Here is a summary:

New endpoints
- Create:  `POST /api/researchhub_unified_documents/review`
- Update:  `PUT /api/researchhub_unified_documents/review/{review_id}`

The endpoint will create/update review and associated thread (if exists). In the future if we want to have reviews without threads, this will work nicely since threads are not required.

The frontend will call **one endpoint to simplify things**. This way we reduce complexity from the FE and defer it to the backend.

Since we now have two locations that may create threads, I refactored the thread logic [into a service which can be seen here](https://github.com/ResearchHub/researchhub-backend/pull/1063/files#diff-8e6e41d69ec4f4ced73df8f15eb4083eb318e62a676782d50d48f41a62bbda99R10).
I'm familiar with this pattern in other frameworks. It seems to be good practice for Django [as described here](https://stackoverflow.com/questions/12578908/separation-of-business-logic-and-data-access-in-django/12579490#12579490)

It makes reading the associated view endpoints much more easily. 
 https://github.com/ResearchHub/researchhub-backend/pull/1063/files#diff-fa1c0e8aa2b9a59ee04c776af8ba75ec965f78fe052a521a833053176d5b2d53R47

le-sun (changes requested):
Is this rebased off master?

yattias:
@le-sun not yet

yattias:
Now rebased

src/discussion/services/thread_service.py
```diff
@@ -0,0 +1,42 @@
...
+                unified_document.id,
+                thread.id,
+            ),
+            priority=1,
+            countdown=10
+        )
+
+    return thread
+
+def update_thread(data, context):
```
yattias, line 37: WIP

src/discussion/serializers.py
```diff
@@ -601,6 +626,7 @@ def get_comments(self, obj):
         return comment_serializer.data
 
     def get_comment_count(self, obj):
+        print('obj', obj)
```
le-sun, line 629: Delete?

src/review/views/review_view.py
```diff
@@ -0,0 +1,130 @@
...
+        IsAuthenticatedOrReadOnly
+        & CreateReview
+        & UpdateReview
+    ]
+    filter_backends = (OrderingFilter,)
+    order_fields = '__all__'
+    ordering = ('-created_date',)
+
+    def create(self, request, pk, **kwargs):
+        unified_document = ResearchhubUnifiedDocument.objects.get(id=pk)
```
le-sun, line 48: I don't think the create function has a pk argument.

src/review/permissions.py
```diff
@@ -0,0 +1,16 @@
+from utils.permissions import (
+    AuthorizationBasedPermission,
+    RuleBasedPermission,
+)
+
+class CreateReview(RuleBasedPermission):
+    message = 'Not allowed to create a review'
+
+    def satisfies_rule(self, request):
+        return True
```
le-sun, line 10: ?

le-sun (changes requested):

PR #1066, opened 2022-04-21
API Key

le-sun:
Added package to handle main API key generation
Implemented CRUD operations for API keys
Updated paper permissions to allow for API keys

Fixed some minor bugs


src/researchhub/middleware/ApiTokenSession.py
```diff
@@ -0,0 +1,30 @@
...
+    from user.models import UserApiToken
+
+    key_parser = KeyParser()
+    user = None
+    key = key_parser.get(request)
+    try:
+        token = UserApiToken.objects.get_from_key(key)
+        user = token.user
+    except UserApiToken.DoesNotExist:
+        user = AnonymousUser()
```
yattias, line 18: Do we want to allow anonymous user here?
Does this mean anonymous user will be able to create papers?

src/user/permissions.py
```diff
@@ -1,28 +1,58 @@
...
+            return True
+
+        return False
+
+    def has_permission(self, request, view):
+        user = request.user
+        if user.is_anonymous:
+            return False
+
+        if request.method in (POST, GET, DELETE):
```
yattias, line 55: Does `GET` make sense here?

src/user/permissions.py
```diff
@@ -1,28 +1,58 @@
...
+    message = "Need to be a moderator."
 
     def has_permission(self, request, view):
         return request.user.moderator
 
     def is_authorized(self, request, view, obj):
         return request.user.moderator
+
+
+class CreateOrViweOrRevokeUserApiToken(BasePermission):
```
yattias, line 37: Typo in View

yattias (changes requested):
Code looks really good.
- Had a few questions (not necessarily blocking)
- Can we write some tests mainly to validate permissions around API tokens work as expected?

src/researchhub/middleware/ApiTokenSession.py
le-sun: It is okay to have anonymous users here, as it is a part of the whole Django auth process, so it will eventually get caught down the line - meaning that anonymous users won't be able to upload papers because they aren't authenticated. But it probably is better to raise an exception here instead. Good catch.

src/user/permissions.py
le-sun: Yep, this is here so that users can see if they've made any tokens.

le-sun:
> Code looks really good.
> 
> * Had a few questions (not necessarily blocking)
> * Can we write some tests mainly to validate permissions around API tokens work as expected?

Added in a test for uploading a paper via token

calvinhlee23:
lgtm. Let's merge this

PR #1068, opened 2022-04-27
Allowing Hub Editors to Remove Discussions

le-sun:
Created a permission that allows any hub editor to remove discussions

### DIFF
--- a/src/discussion/permissions.py
+++ b/src/discussion/permissions.py
@@ -1,145 +1,154 @@
+from rest_framework.permissions import BasePermission
+
+from hypothesis.models import Citation
 from hypothesis.related_models.hypothesis import Hypothesis
-from researchhub.lib import get_document_id_from_path
 from paper.models import Paper
+from researchhub.lib import get_document_id_from_path
 from user.models import Author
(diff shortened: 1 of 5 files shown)


src/discussion/permissions.py
```diff
@@ -1,145 +1,154 @@
...
 from utils.permissions import (
     AuthorizationBasedPermission,
+    PermissionDenied,
     RuleBasedPermission,
-    PermissionDenied
 )
-from hypothesis.models import Citation
+
+
+class EditorCensorDiscussion(BasePermission):
```
le-sun, line 15: New permission

yattias (changes requested):
I requested changes because tests are failing. Could we also add tests ensuring that editors can remove and non-editors cannot?

Other than that, looks good!

le-sun:
> I requested changes because tests are failing. Could we also add tests ensuring that editors can remove and non-editors cannot?
> 
> Other than that, looks good!

Since allowing editors to remove comments is a temporary fix, I think tests aren't really necessary because once we have a solution for spammers, we will remove this permission.

PR #1069, opened 2022-04-28
Peer Reviews

yattias:
Solves #1062


src/discussion/reaction_views.py
```diff
@@ -110,7 +110,34 @@ def censor(self, request, pk=None):
         Contribution.objects.filter(
             content_type=content_type, object_id=item.id
         ).delete()
-        return Response(self.get_serializer(instance=item).data, status=200)
+
+        try:
```
yattias, line 114: If review is associated with this thread, we want to clear caches.

src/discussion/serializers.py
```diff
@@ -164,9 +166,27 @@ def get_post(self, thread):
...
+                obj.review,
+                context=context,
+                **_context_fields
+            )
+
+            return serializer.data
+
+        return None
+
     def get_peer_review(self, obj):
```
yattias, line 184: Note that we  will need to purge/refactor codebase of peer_review once we figure out next steps

src/discussion/views.py
```diff
@@ -113,6 +114,12 @@ def create(self, request, *args, **kwargs):
 
         response = super().create(request, *args, **kwargs)
         response = self.get_self_upvote_response(request, response, Thread)
+
+        created_thread = Thread.objects.get(id=response.data['id'])
```
yattias, line 118: By default, `response` is a flat object without the nested attributed that we need on the frontend (in this case, `reviews`) Rather than making another call to get this data, I am using the serializer class to return properly serialized data (see return statement).

This way, all of our requests to the BE return a consistent object. ATM, creating a thread returns a different obj than fetching threads.

src/hypothesis/serializers/hypothesis_serializer.py
```diff
@@ -59,6 +60,7 @@ class Meta:
             'slug',
             'src',
             'unified_document',
+            'unified_document_id',
```
yattias, line 63: When creating reviews on  the FE, posts and papers have   `unified_document_id` but hypothesis does not. This leads to inconsistency.

src/discussion/views.py
yattias: I verified creating threads, comments, responses works as expected.

src/hypothesis/serializers/hypothesis_serializer.py
le-sun: I agree with the consistency issue. I'm not sure why but the posts and papers serializer includes `unified_document_id`, when `unified_document` works. I think we should get rid of `unified_document_id` and and change the other serializers to use `unified_document` instead.

src/paper/serializers.py
```diff
@@ -267,6 +268,11 @@ def get_user_vote(self, paper):
                     pass
         return vote
 
+    def get_unified_document(self, paper):
+        from researchhub_document.serializers.researchhub_unified_document_serializer import MinimalUnifiedDocumentSerializer
+        uni_doc = MinimalUnifiedDocumentSerializer(paper.unified_document)
+        return uni_doc.data
+
```
le-sun, line 275: I think using a dynamic serializer here is better than defining a new serializer.

src/researchhub_document/serializers/researchhub_unified_document_serializer.py
```diff
@@ -74,6 +74,22 @@ def get_documents(self, instance):
...
+            'id',
+            'reviews'
+        ]
+        read_only_fields = [
+            'id',
+            'reviews',
+        ]
+
+    def get_reviews(self, obj):
+        return obj.get_review_details()
```
le-sun, line 92: I don't think this serializer is necessary.

src/review/serializers/review_serializer.py
```diff
@@ -0,0 +1,26 @@
...
+            'created_date',
+            'updated_date',
+        ]
+
+class DynamicReviewSerializer(
+    DynamicModelFieldSerializer,
+):
+    class Meta:
+        model = Review
+        fields = '__all__'
```
le-sun, line 26: Requires dynamic methods for `created_by` and `unified_document`

src/discussion/serializers.py
```diff
@@ -553,6 +577,7 @@ class Meta:
             'title',
             'user_flag',
             'user_vote',
+            'unified_document',
```
le-sun, line 580: See comment on line 661.

src/discussion/serializers.py
```diff
@@ -627,6 +658,11 @@ def get_peer_review(self, obj):
 
         return None
 
+    def get_unified_document(self, obj):
+        from researchhub_document.serializers.researchhub_unified_document_serializer import MinimalUnifiedDocumentSerializer
+        uni_doc = MinimalUnifiedDocumentSerializer(obj.unified_document)
+        return uni_doc.data
+
```
le-sun, line 665: What is the purpose of having `unified_document` here? Serializer bloat can become a serious issue when it comes to performance.

If it is necessary, I think using a dynamic serializer here is better than defining a new serializer.

le-sun (changes requested):

src/paper/serializers.py
yattias: Yeah sounds reasonable

le-sun (approved):

PR #1076, opened 2022-05-06
[GenericReaction][Migration] - Flag Reasons & Paper Migration

calvinhlee23:
### Planned Migrations Steps: 
- Rename Paper's Vote & Flag to VotePaperLegacy & FlagPaperLagacy
- Attach GRM (generic reaction model) to Paper
- Migrate (1) legacy data to GRM data.
- Update Views & endpoint in both FE & BE

### Test: 
Screen Shot 2022-05-06 at 2 36 14 PM


src/utils/models.py
```diff
@@ -1,11 +1,12 @@
...
 from django.utils import timezone
+
 from utils.managers import SoftDeletableManager
 
 
 class DefaultModel(models.Model):
-    created_date = models.DateTimeField(auto_now_add=True)
-    updated_date = models.DateTimeField(auto_now=True)
+    created_date = models.DateTimeField(auto_now_add=True, null=True)
+    updated_date = models.DateTimeField(auto_now=True, null=True)
```
le-sun, line 9: I think dates should not be able to be null

le-sun (changes requested):
Left a comment, otherwise most things look ok. Test cases are also failing.

src/utils/models.py
le-sun: This is adding a lot of unnecessary migrations too.

PR #1079, opened 2022-05-11
[GenericReaction][Migration] - Flag Reasons & Paper Migration 

calvinhlee23:
Screen Shot 2022-05-10 at 3 12 06 PM

Screen Shot 2022-05-10 at 6 48 45 PM


src/invite/migrations/0007_auto_20220510_2207.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.2 on 2022-05-10 22:07
+
+from django.db import migrations, models
```
yattias, line 3: Are these migrations okay? Not sure why there are note related migrations

yattias (approved):
Looks good except one typo. We should test on staging.

PR #1082, opened 2022-05-12
Paper migration Uploaded_date to Created_date

calvinhlee23:
- Paper migration Uploaded_date to Created_date
- added uploaded_date as @property


le-sun (changes requested):
Looks good. I see that there are some places in the paper_views and researchhub_post_model that uses uploaded date. Is it safe to leave it as is?

le-sun (approved):

PR #1087, opened 2022-05-25
Audit Dashboard Performance

le-sun:
Minor bug fixes: Calculating paper hot score and user authored papers

Added extra indexes to action model and flag model
Updated queries


yattias (approved):
LGTM

PR #1089, opened 2022-05-26
Admin button

yattias:


src/paper/views/paper_views.py
```diff
@@ -370,32 +370,6 @@ def censor(self, request, pk=None):
 
         return Response("Paper was deleted.", status=200)
 
-    @action(
```
yattias, line 373: censor on unified doc is used instead

le-sun (changes requested):
Can you run `pre-commit install`? I don't see the automatic Black formatting happening.

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -634,6 +655,54 @@ def compare(doc):
             }
             return Response(res, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=["post"], permission_classes=[AllowAny])
+    def exclude_from_feed(self, request, pk=None):
+        unified_document = self.queryset.get(id=pk)
```
le-sun, line 660:
```suggestion
        unified_document = self.get_object()
```

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -634,6 +655,54 @@ def compare(doc):
             }
             return Response(res, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=["post"], permission_classes=[AllowAny])
```
le-sun, line 658: Permission class update?

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -634,6 +655,54 @@ def compare(doc):
...
+        reset_unified_document_cache(
+            hub_ids,
+            document_type=["all", doc_type],
+            filters=[TRENDING],
+            with_default_hub=True,
+        )
+
+        return Response(status=status.HTTP_200_OK)
+
+    @action(detail=True, methods=["post"], permission_classes=[AllowAny])
```
le-sun, line 685: Permission class update?

src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -78,7 +88,7 @@ def censor(self, request, pk=None):
             filters=[NEWEST, TOP, TRENDING, DISCUSSED],
             with_default_hub=True,
         )
-
+        print("self.get_serializer(instance=doc)", self.get_serializer(instance=doc))
```
le-sun, line 91: Is this print statement necessary?

src/researchhub_document/related_models/feed_exclusion_model.py
```diff
@@ -0,0 +1,24 @@
...
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+from utils.models import DefaultModel
+
+
+class FeedExclusion(DefaultModel):
+    # Not a foreign key because hub_id=0 is homepage
+    hub_id = models.IntegerField(
+        default=False,
```
le-sun, line 13: Should the default here be false or an int?

src/researchhub_document/related_models/feed_exclusion_model.py
```diff
@@ -0,0 +1,24 @@
...
+        db_index=True,
+    )
+
+    unified_document = models.ForeignKey(
+        ResearchhubUnifiedDocument,
+        default=False,
```
le-sun, line 19: I don't think ForeignKeys have a default

src/researchhub_document/related_models/feed_exclusion_model.py
```diff
@@ -0,0 +1,24 @@
...
+        blank=False,
+        null=False,
```
le-sun, line 21: These two should be True?

le-sun (changes requested):

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: Good catch Leo 👍

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: Good catch @le-sun

src/researchhub_document/related_models/feed_exclusion_model.py
yattias: Apparently they do, I just checked. I think this would just indicate that this field will never have a default value

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: nope

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: @le-sun good catch

src/researchhub_document/related_models/feed_exclusion_model.py
yattias: I think they should be false. I never want this value to be null or blank since I'm excluding unified documents from a particular feed.

le-sun: Oh yeah you're correct. I mixed it up in my head.

src/researchhub_document/related_models/feed_exclusion_model.py
le-sun: Feels weird to have it included, since it will never have a default value unless we add one in, but it's probably ok.

le-sun (approved):
Looks good

PR #1091, opened 2022-05-27
[ContentModeration] - Flagging & Notification 

calvinhlee23:
Screen Shot 2022-05-26 at 5 28 46 PM
Screen Shot 2022-05-26 at 5 28 55 PM

### DIFF
--- a/src/user/views/audit_views.py
+++ b/src/user/views/audit_views.py
@@ -1,3 +1,4 @@
+from django.contrib.admin.options import get_content_type_for_model
 from django.contrib.contenttypes.models import ContentType
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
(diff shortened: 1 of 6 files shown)


src/user/views/audit_views.py
```diff
@@ -249,11 +251,11 @@ def contributions(self, request):
 
     @action(detail=False, methods=["post"])
     def flag(self, request):
-        moderator = request.user
+        hub_editor = request.user
```
yattias, line 254: I understand that `editor` is a better name than `moderator` here but it could be that a moderator who is not an editor can also flag. Maybe we just say user?

But then again, what is this endpoint used for? If it is used for flagging then we should ensure anyone can flag and not just editors or moderators

src/user/views/audit_views.py
```diff
@@ -263,13 +265,13 @@ def flag(self, request):
 
     @action(detail=False, methods=["post"])
     def flag_and_remove(self, request):
-        moderator = request.user
+        hub_editor = request.user
```
yattias, line 268: same as above

src/user/views/audit_views.py
```diff
@@ -303,11 +308,11 @@ def flag_and_remove(self, request):
 
     @action(detail=False, methods=["post"])
     def dismiss_flagged_content(self, request):
-        moderator = request.user
+        hub_editor = request.user
```
yattias, line 311: same above

lightninglu10:
Are users here getting an email in addition to an in app notif?

PR #1093, opened 2022-06-02
OpenAlex Integration

le-sun:
Added OpenAlex to paper upload flow (url/doi)
Added `is_open_access` field to paper model

Minor bug fixes
- Fixed email generation for supporting posts
- Fixed leaderboard ordering for papers


src/paper/management/commands/backfill_openalex.py
```diff
@@ -0,0 +1,45 @@
...
+                pdf_license_url = host_venue.get("url", None)
+                external_source = host_venue.get("display_name", None)
+
+                if pdf_license:
+                    paper.pdf_license = pdf_license
+                if pdf_license_url:
+                    paper.pdf_license_url = pdf_license_url
+                if external_source:
+                    paper.external_source = external_source
+                paper.save()
```
yattias, line 41: This can cause a bunch of other caches to clear - which I think should be fine but needs to be monitored.
Perhaps we should start with backfilling a couple of weeks to see the amount of work added to celery.

If you want to make the start date configurable, you can look at my `recalc_hot_score.py` command I wrote a while back.

PR #1096, opened 2022-06-03
Paper Flag Grm Backfill Script

calvinhlee23:
Screen Shot 2022-06-03 at 4 11 17 PM

### DIFF
--- a/src/discussion/management/commands/backfill_paper_to_grm.py
+++ b/src/discussion/management/commands/backfill_paper_to_grm.py
@@ -15,7 +15,6 @@ class Command(BaseCommand):
         papers = Paper.objects.filter(
             created_date__lte=today,
             created_date__gte=paper_sync_stop_date,
-            unified_document__isnull=True,
         ).order_by("created_date")
         count = papers.count()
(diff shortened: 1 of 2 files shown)


src/discussion/management/commands/backfill_paper_to_grm.py
```diff
@@ -0,0 +1,34 @@
...
+            for j, vote_legacy in enumerate(legacy_votes.iterator()):
+                try:
+                    grm_vote = GrmVote(
+                        created_by=vote_legacy.created_by,
+                        created_date=vote_legacy.created_date,
+                        item=paper,
+                        updated_date=vote_legacy.created_date,
+                        vote_type=vote_legacy.vote_type,
+                    )
+                    grm_vote.save()
```
le-sun, line 32: I think this might be slightly better?
```suggestion
from paper.models import Vote as PaperVote

for i, paper_vote in enumerate(PaperVote.objects.all()):
    ...
    GrmVote.objects.create(
        ...
        item=paper_vote.paper
    )
```

le-sun (changes requested):

le-sun (approved):

PR #1100, opened 2022-06-09
Featured docs

yattias:
- Adding ability for moderator to feature documents in either the homepage, or hub homepage (Trending only)


src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -475,8 +476,21 @@ def get_filtered_queryset(
...
+                "unified_document"
+            )
+
+            if featured_doc_ids.count() > 0:
+                featured_docs = ResearchhubUnifiedDocument.objects.filter(
+                    id__in=featured_doc_ids
+                )
+                qs = qs.exclude(id__in=featured_docs.values_list("id", flat=True))
+                qs = qs.order_by("-hot_score_v2")
+                qs = list(featured_docs) + list(qs)
```
yattias, line 490: We want to preserve the ordering of the original queryset and append `featured_docs` to the front. After multiple attempts with `union` and `|`, this is the only thing that worked. Confirmed that pagination and result set is proper.

Answer borrowed from https://stackoverflow.com/questions/18235419/how-to-chain-django-querysets-preserving-individual-order

le-sun (approved):

src/researchhub_document/views/researchhub_unified_document_views.py
yattias: Thanks @le-sun, I will benchmark this and see what queries this thing generates.
LMK if you have any other ideas here.

src/researchhub_document/related_models/featured_content_model.py
```diff
@@ -0,0 +1,25 @@
...
+
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+from utils.models import DefaultModel
+
+
+class FeaturedContent(DefaultModel):
+    # Not a foreign key because hub_id=0 is homepage
+    hub_id = models.IntegerField(
```
lightninglu10, line 12: I think this needs to be a foreign key and if it's null then it will be on the homepage

lightninglu10 (changes requested):

src/researchhub_document/related_models/featured_content_model.py
lightninglu10: I also think in here, we should add something like so:

```
featured = models.BooleanField(...)
```

this way we can do a soft delete and not a hard delete when we remove from featured

yattias: That's a good suggestion

PR #1105, opened 2022-06-17
PDF Download Update

le-sun:
Updated logic for downloading pdfs
- Updated helper methods to use cloudscrapper to bypass some blocking websites

Adding in OpenAlex open access url to papers, if it exists


yattias (approved):
LGTM

PR #1122, opened 2022-07-20
Send notification to bounty creators after 29 days

lightninglu10:
Send a notification to bounty creators after 29 days notifying them their bounty will expire


src/reputation/tasks.py
```diff
@@ -18,17 +20,76 @@
...
+@periodic_task(
+    run_every=crontab(minute="0", hour="10"),
+    priority=3,
+    queue=QUEUE_BOUNTY_NOTIFICATIONS,
+)
+def send_bounty_notifications():
+    twentynine_days_ago = start_date = datetime.datetime.now() - datetime.timedelta(29)
+    bounties = Bounty.objects.filter(
+        status=Bounty.OPEN,
+        created_date__date=twentynine_days_ago,
```
yattias, line 57: Is it possible to add a condition that examines if the user received a notification already? Reason being is, if we tweak the filtering logic, it could result in user getting more than one notification or not getting any at all.

yattias (changes requested):
Thanks for doing this, looks good with the exception of one concern I had.

Also, mind sharing a screenshot of the email?

PR #1127, opened 2022-07-21
Accepted answer

calvinhlee23:
see title


src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -192,17 +192,17 @@ def _get_serializer_context(self):
...
                     "abstract",
                     "aggregate_citation_consensus",
                     "created_by",
                     "created_date",
+                    "discussion_count",
+                    "featured",
                     "file",
                     "first_preview",
-                    "hot_score",
+                    "has_accepted_answer" "hot_score",
```
le-sun, line 203: Missing comma

le-sun (changes requested):

src/discussion/views.py
```diff
@@ -226,6 +227,30 @@ def upvote(self, *args, **kwargs):
...
+        try:
+            document_id = get_document_id_from_path(self.request)
+            target_post_question = ResearchhubPost.objects.get(id=document_id)
+
+            # logical ordering - DO NOT CHANGE THE ORDER OF OPERATIONS
+            target_thread = self.get_object()
+            target_thread.is_accepted_answer = True
+            target_thread.save()
+
+            current_accepted_answer = target_post_question.get_accepted_answer()
```
yattias, line 245: Can we change to `prev_accepted_answer`? `current` is confusing.

I think since we save the other already, it is okay to consider it as `prev`

PR #1129, opened 2022-08-01
Anonymizing Editor Flagging/Removal of Content

le-sun:
image
image

### DIFF
--- a/src/user/views/audit_views.py
+++ b/src/user/views/audit_views.py
@@ -14,7 +14,7 @@ from mailing_list.lib import base_email_context
 from notification.models import Notification
 from paper.related_models.paper_model import Paper
 from user.filters import AuditDashboardFilterBackend
-from user.models import Action
+from user.models import Action, User
 from user.permissions import IsModerator, UserIsEditor
 from user.serializers import DynamicActionSerializer, VerdictSerializer
 from utils import sentry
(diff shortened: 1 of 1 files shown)


src/user/views/audit_views.py
```diff
@@ -450,3 +452,15 @@ def _send_email_notification_to_content_creator(self, flag, notification, verdic
             "flagged_and_removed_content.html",
             "ResearchHub Digest <digest@researchhub.com>",
         )
+
+    def _get_anonymous_remover(self):
+        user = User.objects.filter(email="community@researchhub.com")
+        if user.exists():
+            return user.first()
+
+        user = User.objects.filter(email="bank@researchhub.com")
```
lightninglu10, line 461: why do we have this as an anonymous remover?

le-sun: The data will return the editor who removed the notification. Since the action object contains that information, we can safely remove it from the notification object.

yattias (approved):

src/user/views/audit_views.py
lightninglu10: I guess I'm not understanding the mechanism why we're doing user.first()

le-sun: This is a fail safe for dev environments since not all local environments has an "anonymous user"

le-sun: This is a fail safe for dev environments since not all local environments has an "anonymous user"

PR #1141, opened 2022-09-06
Boost bounties in feed

yattias:
Promote docs with bounties if they fall within the "near new" or "near expire" zones. These zones are basically three day segments near create data and expire date

image


le-sun:
Tests are failing

PR #1145, opened 2022-09-14
Minor Bugfixes

le-sun:
Fixed emails sent by supporting documents with RSC
Updated notification method to serialize user actions
Added temporary email check for email notifications


yattias (approved):
LGTM

PR #1150, opened 2022-09-25
Add tag app; Concept model many-to-many w/ Paper

vlazzle:
more context in [the Content Grouping doc](https://docs.google.com/document/d/16FDcywdc3kjERkAkA73jaLrpcJoGb30lYZYinKGpDuQ/edit)


src/hub/views.py
```diff
@@ -235,92 +235,6 @@ def invite_to_hub(self, request, pk=None):
 
         return Response(response, status=200)
 
-    @action(detail=True, methods=[GET])
-    def latest_actions(self, request, pk=None):
```
yattias, line 239: Not sure why this shows as removed in your branch. I think it has to do with some work we recently merged unless you also did related work on this? Not sure if this change is related to your branch

src/reputation/migrations/0062_auto_20220926_1846.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2 on 2022-09-26 18:46
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
yattias, line 6: Is this migration also related to your branch? Seems like it may not

src/tag/migrations/0001_initial.py
```diff
@@ -0,0 +1,30 @@
...
+            name='Concept',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_date', models.DateTimeField(auto_now_add=True)),
+                ('updated_date', models.DateTimeField(auto_now=True)),
+                ('openalex_id', models.CharField(max_length=255)),
+                ('display_name', models.CharField(max_length=255)),
+                ('description', models.TextField(blank=True, default='')),
+                ('openalex_created_date', models.CharField(max_length=255)),
+                ('openalex_updated_date', models.CharField(max_length=255)),
```
yattias, line 24: I don't think we should store the `openalex_created_date` and `openalex_updated_date` because they don't give us value (though I could be missing something) also, openalex concept may be updated on their end but the `openalex_updated_date` will not reflect this change because they do not push changes to us.

src/tag/models.py
```diff
@@ -0,0 +1,38 @@
+from django.db import models
+from utils.models import DefaultModel
+
+# We ask openalex to classify a paper into concepts, which we then present as tag candidates from which users can pick zero or more to attach to the paper.
```
yattias, line 4: Are we going to ask users to pick tags? Perhaps we shouldn't do this because openalex already did so.
I think we should just accept and persist the list of concepts - at least for V1

src/reputation/migrations/0062_auto_20220926_1846.py
vlazzle: You're right it's not, it came from https://github.com/ResearchHub/researchhub-backend/pull/1151 and I think it's appearing int his PR because I merged in master after pushing this PR initially and pushed again.

This is what I had in mind when I said "It looks a bit messy to view but I found that you can pick my commits in the Files changed tab so it doesn’t show diffs for the stuff merged in from master." Alternatively, as the owner of this PR, I can save you that trouble if I rebase my commits on top of master and force push to this PR.

yattias (commented):
@vlazzle the code looks good but I think we should make a change

I think concepts should be associated with `unified_document` (which is out generic document object) as opposed to papers. That way, we can apply concepts to questions, posts, ... in addition to just papers. I think the change here shouldn't be too time consuming to make. @le-sun / @lightninglu10 lmk if this makes sense

Also, I think it may make sense to open a new PR with just the changes. I'm not sure why I'm seeing what appears to be unrelated diffs in your PR (see my comments). Maybe something happened during rebase?

le-sun:
> @vlazzle the code looks good but I think we should make a change
> 
> I think concepts should be associated with `unified_document` (which is out generic document object) as opposed to papers. That way, we can apply concepts to questions, posts, ... in addition to just papers. I think the change here shouldn't be too time consuming to make. @le-sun / @lightninglu10 lmk if this makes sense
> 
> Also, I think it may make sense to open a new PR with just the changes. I'm not sure why I'm seeing what appears to be unrelated diffs in your PR (see my comments). Maybe something happened during rebase?

I agree. Having concepts under unified document makes it more generic.

src/reputation/migrations/0062_auto_20220926_1846.py
le-sun: I think rebasing on master and force pushing is good.

lightninglu10: agree with rebasing on top of master and force pushing, this way we know exactly what's inside of the PR

src/tag/migrations/0001_initial.py
lightninglu10: i think it's OK to have this data, in case we ever want to do anything with it

src/tag/models.py
lightninglu10: in the user workflow, we'll have users pick which tags they should accept from OpenAlex. Discussed in the google doc

src/tag/models.py
```diff
@@ -0,0 +1,38 @@
...
+# TODO: migration
+# https://docs.openalex.org/about-the-data/concept
+class Concept(DefaultModel):
```
lightninglu10, line 7: Instead of calling this Concept (kind of generic), can we name this to something else? Like OpenAlexConcept

src/tag/models.py
vlazzle: Users picking tags from the concepts openalex returns is what we agreed on in the [doc](https://docs.google.com/document/d/16FDcywdc3kjERkAkA73jaLrpcJoGb30lYZYinKGpDuQ/edit), but we may need to cut it from the trial project scope. I'd love to get it in but will have a better idea on how realistic that is as we approach the presentation on 10/14.

src/tag/migrations/0001_initial.py
vlazzle: The value I was thinking of is mainly to see if our version of the concept is outdated by comparing against the same concept's updated_date as returned by the openalex api.

Also it may come in handy for debugging purposes later.

vlazzle:
Closing this, let's continue in #1155

src/reputation/migrations/0062_auto_20220926_1846.py
vlazzle: ok, i'll do that henceforth. thanks!

PR #1157, opened 2022-09-30
get dehydrated concepts from openalex, then sort them by level & score

vlazzle:
Use the id from each dehydrated concept returned from OpenAlex's /works endpoint to fetch hydrated concepts from the /concepts endpoint.

Refactor OpenAlex class to share code between _get_works and get_hydrated_concept.


yattias (approved):
Looks good

PR #1158, opened 2022-10-01
Associate Concepts to a Paper through unified_document

vlazzle:
1. check local db for hydrated concepts and skip openalex if local cache is "fresh"
2. update stored concepts with data from openalex if they need a refresh
3. create model objects from hydrated concepts
4. associate concept model objects to a paper through unified_document

also:
- make Concept.openalex_id a unique key.
- catch http error from hydrate_concept & proceed w/ celery_openalex sans concepts.


yattias (approved):
Looks good

PR #1167, opened 2022-10-08
Added Twitter config and updated README

AaronWatson2975:
# Description
When I pulled the repo and attempted to run the backend using the steps included in the readme, I hit the following error.

```
ImportError: cannot import name 'twitter' from 'config_local' (/usr/app/src/config_local/__init__.py)
```

To address this I looked in `src/researchhub/settings.py`, copied the keys into a new config file following the same pattern as the other two, and updated the README to reflect the change.

# Testing Notes
### Procedure
Testing this is fairly simple.
1. Clone a fresh copy of the repository
2. Follow the steps under "Quick install using Docker (Recommended)"

### Expected Result
The backend should build and run without any blocking errors.


twitter_config_sample.py
```diff
@@ -0,0 +1,12 @@
+TWITER_ACCESS_TOKEN="PLACEHOLDER"
+TWITER_ACCESS_TOKEN_ALT="PLACEHOLDER"
```
AaronWatson2975, line 2: I did notice this typo when I was in here but for the sake of not breaking everyones local developer environments and going down a rabbit hole, I left it as is.

le-sun:
Looks good, thanks!

PR #1168, opened 2022-10-10
adding should_display_rsc_balance_home

calvinhlee23:


le-sun:
Tests are failing

src/user/related_models/user_model.py
```diff
@@ -35,28 +35,29 @@ def editors(self):
...
-    invited_by = models.ForeignKey(
-        "self", related_name="invitee", on_delete=models.SET_NULL, null=True, blank=True
-    )
+    reputation = models.IntegerField(default=100)
+    should_display_rsc_balance_home = models.BooleanField(default=True)
     sift_risk_score = models.FloatField(null=True, blank=True)
+    spam_updated_date = models.DateTimeField(null=True)
+    suspended_updated_date = models.DateTimeField(null=True)
+    updated_date = models.DateTimeField(auto_now=True)
+    upload_tutorial_complete = models.BooleanField(default=False)
```
le-sun, line 60: Thanks for sorting

src/user/related_models/user_model.py
```diff
@@ -35,28 +35,29 @@ def editors(self):
...
+    moderator = models.BooleanField(default=False)
     probable_spammer = models.BooleanField(default=False)
-    suspended_updated_date = models.DateTimeField(null=True)
-    spam_updated_date = models.DateTimeField(null=True)
     referral_code = models.CharField(max_length=36, default=uuid.uuid4, unique=True)
-    invited_by = models.ForeignKey(
-        "self", related_name="invitee", on_delete=models.SET_NULL, null=True, blank=True
-    )
+    reputation = models.IntegerField(default=100)
+    should_display_rsc_balance_home = models.BooleanField(default=True)
```
le-sun, line 55: I like your idea of separating out user preferences into another model, but this is fine for now

src/user/views/user_views.py
```diff
@@ -124,6 +124,29 @@ def referral_rsc(self, request):
...
         return Response({"amount": amount})
 
+    @action(
+        detail=True,
+        methods=[RequestMethods.POST],
+        permission_classes=[RequestorIsOwnUser],
+    )
+    def set_should_display_rsc_balance(self, request, pk=None):
+        try:
+            user = User.objects.get(id=request.user.id)
```
le-sun, line 134: Is the request coming as `api/user/{user_id}/set_should_display_rsc_balance`? If so, it can be `user = self.get_object()`

src/user/views/user_views.py
```diff
@@ -124,6 +124,29 @@ def referral_rsc(self, request):
...
+            target_value = request.data.get("should_display_rsc_balance_home")
```
le-sun, line 135: Should we have some input check here? To ensure that it is a bool

le-sun (changes requested):

src/user/views/user_views.py
calvinhlee23: i think it's ok as is. If it's not a bool, it will throw an error because of the model field type.

src/user/views/user_views.py
calvinhlee23: i want the FE payload to specify which user is the target user for the endpoint. 
Permission check is already being done.

PR #1171, opened 2022-10-12
Hotwallet low balance alert

lightninglu10:
Sending emails for whenever the hotwallet has low balance

### DIFF
--- a/src/reputation/lib.py
+++ b/src/reputation/lib.py
@@ -414,17 +414,18 @@ def check_hotwallet():
         outer_subject = "ETH is running low in the hotwallet"
         send_email = True
 
-    # if send_email:
-
-    context = {**base_email_context}
-    context["action"] = {
-        "message": "RSC: {:,}\n\nETH: {:,}".format(rsc_balance_eth, eth_balanace_eth)
-    }
-    context["subject"] = outer_subject
-    send_email_message(
-        ["patricklu@researchhub.com", "pat@researchhub.com"],
-        "general_email_message.txt",
-        outer_subject,
-        context,
-        html_template="general_email_message.html",
-    )
+    if send_email:
+        context = {**base_email_context}
+        context["action"] = {
+            "message": "RSC: {:,}\n\nETH: {:,}".format(
+                rsc_balance_eth, eth_balanace_eth
+            )
+        }
+        context["subject"] = outer_subject
+        send_email_message(
+            ["patricklu@researchhub.com", "pat@researchhub.com"],
+            "general_email_message.txt",
+            outer_subject,
+            context,
+            html_template="general_email_message.html",
+        )
--- a/src/reputation/lib.py
+++ b/src/reputation/lib.py
@@ -2,607 +2,337 @@ import json
 import logging
 import os
 
-import ethereum.utils
-import ethereum.lib
 from web3 import Web3
-from utils.sentry import log_error
 
-from ethereum.lib import get_private_key, RSC_CONTRACT_ADDRESS, execute_erc20_transfer
+import ethereum.lib
+import ethereum.utils
+from ethereum.lib import RSC_CONTRACT_ADDRESS, execute_erc20_transfer, get_private_key
+from mailing_list.lib import base_email_context
 from researchhub.settings import (
     ASYNC_SERVICE_HOST,
+    WEB3_KEYSTORE_ADDRESS,
     WEB3_SHARED_SECRET,
     w3,
-    WEB3_KEYSTORE_ADDRESS,
 )
-from utils.http import http_request, RequestMethods
-WITHDRAWAL_MINIMUM = int(os.environ.get('WITHDRAWAL_MINIMUM', 100))
+from utils.http import RequestMethods, http_request
+from utils.message import send_email_message
+from utils.sentry import log_error
+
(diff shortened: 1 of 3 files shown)


src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+
+    contract = w3.eth.contract(
+        abi=contract_abi, address=Web3.toChecksumAddress(RSC_CONTRACT_ADDRESS)
+    )
+    rsc_balance_wei = contract.functions.balanceOf(WEB3_KEYSTORE_ADDRESS).call()
+    decimals = contract.functions.decimals().call()
+    rsc_balance_eth = rsc_balance_wei / (10**decimals)
+    eth_balance_wei = w3.eth.getBalance(WEB3_KEYSTORE_ADDRESS)
+    eth_balanace_eth = eth_balance_wei / (10**18)
+    send_email = False
```
le-sun, line 406: Remove

src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+
+    if rsc_balance_eth <= 300000:
+        outer_subject = "RSC is running low in the hotwallet"
+        send_email = True
+
+    if eth_balanace_eth < 0.08:
+        outer_subject = "ETH is running low in the hotwallet"
+        send_email = True
+
+    # if send_email:
```
le-sun, line 417: Remove

src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+    """
+
+    contract = w3.eth.contract(
+        abi=contract_abi, address=Web3.toChecksumAddress(RSC_CONTRACT_ADDRESS)
+    )
+    rsc_balance_wei = contract.functions.balanceOf(WEB3_KEYSTORE_ADDRESS).call()
+    decimals = contract.functions.decimals().call()
+    rsc_balance_eth = rsc_balance_wei / (10**decimals)
+    eth_balance_wei = w3.eth.getBalance(WEB3_KEYSTORE_ADDRESS)
+    eth_balanace_eth = eth_balance_wei / (10**18)
```
le-sun, line 405:
```suggestion
    eth_balanace_eth = eth_balance_wei / (10**decimals)
```
We can reuse decimals right? I'm assuming 18 == decimals in this case.

src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+    send_email = False
+    outer_subject = "RSC is running low in the hotwallet"
+
+    if rsc_balance_eth <= 300000:
+        outer_subject = "RSC is running low in the hotwallet"
+        send_email = True
+
+    if eth_balanace_eth < 0.08:
+        outer_subject = "ETH is running low in the hotwallet"
+        send_email = True
```
le-sun, line 415: Remove

src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+    decimals = contract.functions.decimals().call()
+    rsc_balance_eth = rsc_balance_wei / (10**decimals)
+    eth_balance_wei = w3.eth.getBalance(WEB3_KEYSTORE_ADDRESS)
+    eth_balanace_eth = eth_balance_wei / (10**18)
+    send_email = False
+    outer_subject = "RSC is running low in the hotwallet"
+
+    if rsc_balance_eth <= 300000:
+        outer_subject = "RSC is running low in the hotwallet"
+        send_email = True
```
le-sun, line 411: Remove

src/reputation/lib.py
```diff
@@ -620,39 +351,80 @@ def __init__(self, withdrawal, balance_record_id, amount):
...
+    if eth_balanace_eth < 0.08:
+        outer_subject = "ETH is running low in the hotwallet"
+        send_email = True
+
+    # if send_email:
+
+    context = {**base_email_context}
+    context["action"] = {
+        "message": "RSC: {:,}\n\nETH: {:,}".format(rsc_balance_eth, eth_balanace_eth)
+    }
```
le-sun, line 422: We might need an additional key `frontend_view_link`, but if the email sends without this, then you can ignore this.

le-sun (changes requested):
Small changes

src/reputation/lib.py
lightninglu10: woops, this should be in there. it needs to be if send_email is true, then send it..

src/reputation/lib.py
lightninglu10: not always the case. decimals is for our contract call, which can be anything. technically it will always be 18 since ours is 18, but i'd like to keep it like this for posterity. eth always has 18 decimals so i stuck it like that

src/reputation/lib.py
lightninglu10: yeah we don't need it in here

le-sun (approved):
Looks good. Can we stage it and try out the email?

lightninglu10:
Screen Shot 2022-10-12 at 3 13 58 PM

Looks like this -- kind of janky but we don't need anything better really for the moment

PR #1172, opened 2022-10-13
RSC Circulating Supply

le-sun:
Created endpoint that displays blockchain and in-app circulating supply (RSC)
Updated periodic tasks to run once every hour from every minute of a specific hour


src/reputation/views/__init__.py
```diff
@@ -27,3 +53,77 @@ def distribute_rsc(request):
...
+    CACHE_KEY = f"{APP_ENV}_RSC_CIRCULATING_SUPPLY"
+
+    results = cache.get(CACHE_KEY)
+    if results:
+        return Response(results, status=200)
+
+    blockchain_supply = get_blockchain_rsc_supply()
+    inapp_supply = get_inapp_rsc_supply()
+    supply = {
+        "RSC Supply": f"{blockchain_supply:,f}",
```
lightninglu10, line 70: can you change this key to eth_blockchain

src/reputation/views/__init__.py
```diff
@@ -27,3 +53,77 @@ def distribute_rsc(request):
...
+        "In-App Supply": f"{inapp_supply:,f}",
```
lightninglu10, line 71: can you change this key to inapp

src/reputation/views/__init__.py
le-sun: Updated

src/reputation/views/__init__.py
le-sun: Updated

PR #1176, opened 2022-10-26
Kobe/referral

yattias:
New Referral program changes.
- Referring users will receive a 7% commission for contribution made by invited user
- Invited user will receive a flat 50RSC upon joining

Invited user logic will be gated behind profile claiming in v2


src/oauth/serializers.py
```diff
@@ -1,30 +1,27 @@
-from allauth.socialaccount.models import SocialAccount
-from allauth.account import app_settings
-from allauth.utils import (
-    get_user_model,
-    email_address_exists
-)
-from rest_framework import serializers
+from time import time
```
yattias, line 1: This file is mostly linting. I'll highlight the actual change below

src/oauth/serializers.py
```diff
@@ -212,45 +196,75 @@ def validate(self, attrs, retry=0):
...
-                user = attrs['user']
-                if referral_code and not user.invited_by and referral_user.id != user.id:
+                user = attrs["user"]
+
+                invited_by_flag_not_set = (
+                    referral_code
+                    and not user.invited_by
+                    and referral_user.id != user.id
+                )
+                if invited_by_flag_not_set:
```
yattias, line 222: This part is new.
This seems to be the best place to pay the invited user bonus.

- Tried using the various allauth signals - request object does not include `referral_code`
- Tried using user `post_save` - but problem is that `invited_by` is set after initial create

Next week, the logic will be moved elsewhere gated by profile claim.

src/reputation/distributor.py
```diff
@@ -84,6 +87,42 @@ def distribute(self):
             print(error_message, e)
         return record
 
+    def _record_referral_distribution_if_applicable(self, original_distribution):
```
yattias, line 90: This part is important. Here we pay a commission to the referer

src/user/signals.py
```diff
@@ -150,45 +150,6 @@ def create_action(sender, instance, created, **kwargs):
                     )
             user = instance.created_by
 
-        """
-        If we're creating an action for the first time,
-        check if we've been referred
-        """
-        referral_content_types = [
```
yattias, line 157: This is old referral program logic where we would pay when invited user created their first contribution

src/user/views/user_views.py
```diff
@@ -72,6 +78,47 @@ def get_serializer_class(self):
     def get_serializer_context(self):
         return {"get_subscribed": True, "get_balance": True, "user": self.request.user}
 
+    @action(detail=False, methods=["GET"], permission_classes=[AllowAny])
```
yattias, line 81: Here we get a list of invited users and the amount of rsc earned by referring them

src/oauth/serializers.py
```diff
@@ -212,45 +192,75 @@ def validate(self, attrs, retry=0):
...
+
+                    has_invited_been_paid = (
+                        Distribution.objects.filter(
+                            distribution_type=REFERRAL_PROGRAM[
+                                "INVITED_DISTRIBUTION_TYPE"
+                            ],
+                            recipient_id=user.id,
+                        ).count()
+                        > 0
+                    )
```
le-sun, line 230:
```suggestion

                    has_invited_been_paid = (
                        Distribution.objects.filter(
                            distribution_type=REFERRAL_PROGRAM[
                                "INVITED_DISTRIBUTION_TYPE"
                            ],
                            recipient_id=user.id,
                        ).exists()
                    )
```

src/reputation/distributor.py
```diff
@@ -71,6 +72,8 @@ def reputation(self):
 
     def distribute(self):
         record = self._record_distribution()
+        self._record_referral_distribution_if_applicable(record)
```
le-sun, line 75: We should move this after line 80, so it only distributes if the original distribution is successful.

src/reputation/distributor.py
le-sun, line 90: We should wrap this in a try catch too.

le-sun (changes requested):

src/user/views/user_views.py
```diff
@@ -72,6 +78,47 @@ def get_serializer_class(self):
...
+        for invited_user in invited:
+            for dist in earned_distributions:
+                if dist["giver_id"] == invited_user["user"]["id"]:
+                    invited_user["rsc_earned"] = dist["rsc_earned"]
+
+            invited_user["benefits_expire_on"] = invited_user[
+                "created_date"
+            ] + relativedelta(months=+6)
+
+        return Response(invited)
```
le-sun, line 120:
```suggestion
    def get_referred_users(self, request):
        invited = User.objects.all().filter(
            invited_by__id=request.user.id
        ).annotate(
            rsc_earned=Sum('reputation_records'), 
            benefits_expire_on=models.ExpressionWrapper(models.F('created_date') + datetime.timedelta(days=30), 
            output_field=models.DateTimeField())
    )
```
Something like this works better. You'll have to modify the serializer and logic a bit, but it is better than iterating through the queryset. You can also add a filter on the `Sum` function to filter by specific distributions.

le-sun (changes requested):
Fix merge conflicts too.

src/reputation/distributor.py
yattias: good call!

src/reputation/distributor.py
yattias: good call

src/user/views/user_views.py
```diff
@@ -72,6 +78,41 @@ def get_serializer_class(self):
...
+                    Q(
+                        reputation_handed_out__distribution_type="REFERRAL_REFERER_EARNINGS"
+                    )
+                    & Q(reputation_handed_out__recipient_id=request.user)
+                ),
+            ),
+            benefits_expire_on=models.ExpressionWrapper(
+                models.F("created_date")
+                + timedelta(
+                    days=REFERRAL_PROGRAM["ELIGIBLE_TIME_PERIOD_IN_MONTHS"] * 30
```
yattias, line 96: @le-sun I can't seem to be able to do this in months. Like `months=6`. Tried using relativedelta but doesn't work here. any ideas?

src/user/serializers.py
```diff
@@ -457,6 +459,12 @@ def get_author_profile(self, user):
             sentry.log_error(e)
             return {}
 
+    def get_rsc_earned(self, user):
+        return user.rsc_earned
```
le-sun, line 463:
```suggestion
        rsc_earned = getattr(user, "rsc_earned", None)
        return rsc_earned
```
This ensures that it won't error out

src/user/serializers.py
```diff
@@ -457,6 +459,12 @@ def get_author_profile(self, user):
...
+
+    def get_benefits_expire_on(self, user):
+        return user.benefits_expire_on
```
le-sun, line 466:
```suggestion
        benefits_expire_on = getattr(user, "benefits_expire_on", None)
        return benefits_expire_on
```
This ensures that it won't error out

le-sun (changes requested):

src/user/serializers.py
yattias: ah good good

src/reputation/views/bounty_view.py
```diff
@@ -53,21 +53,21 @@ class BountyViewSet(viewsets.ModelViewSet):
...
         return User.objects.get(id=1)
 
     def _get_dao_fee_recipient(self):
-        user = User.objects.filter(email="community@researchhub.com")
+        user = User.objects.filter(email="kobe@researchhub.com")
         if user.exists():
             return user.first()
 
-        user = User.objects.filter(email="bank@researchhub.com")
+        user = User.objects.filter(email="kobe@researchhub.com")
```
le-sun, line 70: These values need to be reverted

le-sun (changes requested):

le-sun (approved):
Looks good

PR #1180, opened 2022-11-09
Scholarly

le-sun:
Created external author claim model for Google Scholar results
Added author citation models to track Scholarly results
Implemented default endpoints and Celery task for claiming and adding Google Scholar profiles


src/researchhub_case/constants/case_constants.py
```diff
@@ -28,9 +29,10 @@
 RH_CASE_TYPES = [
     (AUTHOR_CLAIM, AUTHOR_CLAIM),
     (PAPER_CLAIM, PAPER_CLAIM),
+    (EXTERNAL_AUTHOR_CLAIM, EXTERNAL_AUTHOR_CLAIM),
 ]
 
 CLAIM_CONTEXT_ENTITY_TYPES = [
-    ('PAPER', 'PAPER'),
-    ('AUTHOR', 'AUTHOR'),
+    ("PAPER", "PAPER"),
```
yattias, line 36: @le-sun will claims submitted through the paper page be considered "Paper" claims

src/researchhub_case/serializers/external_author_claim_case_serializer.py
```diff
@@ -0,0 +1,32 @@
+from rest_framework.serializers import ModelSerializer
+
+from researchhub_case.models import ExternalAuthorClaimCase
+
+from .abstract_researchhub_case_serializer import DynamicAbstractResearchhubCase
+from .researchhub_case_abstract_serializer import EXPOSABLE_FIELDS
+
+
+class ExternalAuthorClaimCaseSerializer(ModelSerializer):
```
yattias, line 9: Just a naming suggestion, but up to you whether you want to make it.

The way I understand it, the "case" is not "external", but rather the api call to fetch data is.
The case itself is an internal ResearchHub case. For consistency, I would just call it `AuthorClaimCaseSerializer`

yattias (approved):

src/researchhub_case/constants/case_constants.py
le-sun: It will be `EXTERNAL_AUTHOR_CLAIM`. I'll update the list.

src/researchhub_case/related_models/external_author_claim_case_model.py
```diff
@@ -0,0 +1,39 @@
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["google_scholar_id", "requestor"],
+                name="unique_google_scholar_id_claim",
+            )
+        ]
+
+    def approve_google_scholar(self):
+        celery_add_author_citations.apply_async(
```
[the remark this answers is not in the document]
le-sun: A lot of tasks are being called from the model, but specifically in the view. So when an editor approves a claim, it will hit the endpoint and inside the view, it will call this model function.

src/researchhub_case/serializers/external_author_claim_case_serializer.py
le-sun: Yeah, I'd prefer a different name too, but `AuthorClaimCaseSerializer` is already used for claiming author profiles. So, I appended external as a prefix since `external` makes sense for author profiles retrieved from Scholarly (and potentially future external apis)

src/config/keys.py
```diff
@@ -55,3 +55,5 @@
 WEB3_KEYSTORE_ADDRESS = os.environ.get("WEB3_KEYSTORE_ADDRESS", "")
 
 TRANSPOSE_KEY = os.environ.get("TRANSPOSE_KEY", "")
+
+SERP_API_KEY = os.environ.get("SERP_API_KEY", "")
```
lightninglu10, line 59: Do we still need this API key?

le-sun: I can remove it.

PR #1181, opened 2022-11-09
Better pdf pulling for nature and sciencedirect

lightninglu10:
Pulling pdf's for nature and science direct better:

Screen Shot 2022-11-08 at 5 33 26 PM

Screen Shot 2022-11-08 at 5 32 42 PM

### DIFF
--- a/src/paper/related_models/paper_model.py
+++ b/src/paper/related_models/paper_model.py
@@ -776,6 +776,17 @@ class Paper(AbstractGenericReactionModel):
             self.completeness = self.INCOMPLETE
         self.save()
 
+    def download_pdf(self):
+        pass
+
+    def get_pdf_link(self):
+        metadata, converted = populate_pdf_url_from_journal_url(self.url, {})
+        pdf_url = metadata.get("pdf_url")
+        if pdf_url:
+            self.pdf_url = metadata.get("pdf_url")
+            self.save()
+        return metadata, converted
+
 
 class MetadataRetrievalAttempt(models.Model):
     CROSSREF_DOI = "CROSSREF_DOI"
--- a/src/paper/related_models/paper_model.py
+++ b/src/paper/related_models/paper_model.py
@@ -3,6 +3,7 @@ import datetime
 import pytz
 import regex as re
 import requests
+from bs4 import BeautifulSoup
 from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.postgres.fields import JSONField
 from django.contrib.postgres.indexes import HashIndex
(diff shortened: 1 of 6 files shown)


src/utils/http.py
```diff
@@ -83,6 +83,8 @@ def check_url_contains_pdf(url) -> bool:
     if url is None:
         return False
     try:
+        if "sciencedirect" in url and "download=false" in url:
```
lightninglu10, line 86: special clause here -- the sciencedirect doesn't work with the scraper, but the link to their pdf works

src/paper/utils.py
```diff
@@ -128,15 +128,21 @@ def populate_pdf_url_from_journal_url(url, metadata):
...
-            break
+        try:
+            if host in journal_url:
+                if journal_url_to_pdf[host]:
+                    pdf_url = journal_url_to_pdf[host](journal_url)
+                    break
+        except Exception as e:
+            import pdb
+
+            pdb.set_trace()
```
le-sun, line 145: Debugger.

src/paper/related_models/paper_model.py
```diff
@@ -776,6 +776,17 @@ def set_paper_completeness(self):
             self.completeness = self.INCOMPLETE
         self.save()
 
+    def download_pdf(self):
+        pass
```
le-sun, line 780: Is this needed?

src/paper/related_models/paper_model.py
```diff
@@ -776,6 +776,17 @@ def set_paper_completeness(self):
...
+
+    def get_pdf_link(self):
```
le-sun, line 782: Very small change, but we should have an additional arg `should_save=False|True`

src/paper/tasks.py
```diff
@@ -1512,13 +1512,13 @@ def celery_create_paper(self, celery_data):
...
         async_paper_updator = getattr(paper_submission, "async_updator", None)
         paper = Paper(**paper_data)
-
         if async_paper_updator is not None:
             paper.doi = async_paper_updator.doi
             paper.hub.add(*async_paper_updator.hubs)
             paper.title = async_paper_updator.title
 
         paper.full_clean()
+        paper.get_pdf_link()
```
le-sun, line 1521: With the previous suggestion, 
```suggestion
        paper.get_pdf_link(should_save=False)
```
since we are already saving the paper at the end of the method.

le-sun (changes requested):

le-sun (approved):

PR #1185, opened 2022-11-22
Explicitly identify transactions eligible to be paid via referral program

yattias:


src/reputation/distributor.py
```diff
@@ -112,14 +112,23 @@ def _record_referral_distribution_if_applicable(self, original_distribution):
             original_distribution.giver_id
             == original_distribution.recipient.invited_by.id
         )
-        excluded_dist_types = [
-            REFERRAL_PROGRAM["REFERER_DISTRIBUTION_TYPE"],
-            REFERRAL_PROGRAM["INVITED_DISTRIBUTION_TYPE"],
+
+        whitelisted_dist_types = [
```
le-sun, line 116: Could we use the constants in `distributions.py`

le-sun (changes requested):

PR #1188, opened 2022-11-30
time filters

lightninglu10:
Adding in filters to get rep by time

### DIFF
--- a/src/user/serializers.py
+++ b/src/user/serializers.py
@@ -358,11 +358,8 @@ class UserSerializer(ModelSerializer):
             return None
 
     def get_time_rep(self, obj):
-        try:
-            return obj.time_rep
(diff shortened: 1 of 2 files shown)


src/user/serializers.py
```diff
@@ -354,6 +357,12 @@ def get_hub_rep(self, obj):
         except Exception:
             return None
 
+    def get_time_rep(self, obj):
+        try:
+            return obj.time_rep
+        except Exception:
+            return None
+
```
le-sun, line 365:
```suggestion
        time_rep = getattr(obj, "time_rep", None)
        return time_rep
```

le-sun (changes requested):

le-sun (approved):

PR #1190, opened 2022-12-08
Unpaywall Integration

le-sun:
Added in celery task for Unpaywall

Minor fixes -
Updated Github actions
Correctly setting is_removed on action model for verdicts
Adding Vercel to allowed hosts

### DIFF
--- a/src/researchhub/settings.py
+++ b/src/researchhub/settings.py
@@ -54,7 +54,7 @@ if DEVELOPMENT or TESTING:
 elif PRODUCTION:
     BASE_FRONTEND_URL = "https://researchhub.com"
 elif CLOUD:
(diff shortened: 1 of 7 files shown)


src/researchhub/settings.py
```diff
@@ -54,7 +54,7 @@
 elif PRODUCTION:
     BASE_FRONTEND_URL = "https://researchhub.com"
 elif CLOUD:
-    BASE_FRONTEND_URL = "https://staging-web.researchhub.com"
+    BASE_FRONTEND_URL = "https://researchhub-web-staging.vercel.app/"
```
lightninglu10, line 57: can we keep this staging-web?

le-sun: Sure

PR #1196, opened 2023-01-03
Registration with email

yattias:
✅  Safe to release independently of corresponding FE portion

- Allows for registration via email
- Supports routes for changing password, and resetting password
- Email templates


src/user/custom_allauth.py
```diff
@@ -0,0 +1,43 @@
+from allauth.account.adapter import DefaultAccountAdapter, get_adapter
+from allauth.account.forms import ResetPasswordForm
+from allauth.account.utils import user_pk_to_url_str
+from dj_rest_auth.registration.serializers import RegisterSerializer
+from dj_rest_auth.serializers import PasswordResetSerializer
+
+from researchhub.settings import BASE_FRONTEND_URL
+
+
+class CustomAccountAdapter(DefaultAccountAdapter):
```
yattias, line 10: Not sure if there is a more logical place to store these custom override classes

yattias: The following are recommendations from author of DJ-Auth library, needed to override password reset, and change password mechanism

yattias: Goal of which is to use a different frontend url than default

le-sun:
Looks like the login tests are failing. Can you check those?

PR #1200, opened 2023-01-10
[Comment Migration][1/n] - DefaultAuthenticatedModel & AbstractGenericRelationModel

calvinhlee23:
- Model setup


src/utils/models.py
```diff
@@ -12,6 +14,52 @@ class Meta:
...
+        "user.User",
+        blank=False,
+        help_text="Last user to update the instance",
+        null=False,
+        on_delete=models.CASCADE,
+    )
+    updated_date = models.DateTimeField(
+        auto_now_add=True,
+        blank=True,
+        null=False,
```
le-sun, line 42: This will always be populated so we can remove this

src/utils/models.py
```diff
@@ -12,6 +14,52 @@ class Meta:
...
+    created_by = models.ForeignKey(
+        "user.User",
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+    )
+    created_date = models.DateTimeField(
+        auto_now_add=True,
+        blank=True,
+        null=False,
```
le-sun, line 30: This will always be populated so we can remove this

le-sun (changes requested):
Will there be cases where created_by and updated_by will be null?

PR #1201, opened 2023-01-10
[Comment Migration][2/n] - RhCommentModel & RhCommentThreadModel

calvinhlee23:
- models & migrations


src/researchhub_comment/related_models/rh_comment_model.py
```diff
@@ -0,0 +1,55 @@
...
+        auto_created=True,
+        on_delete=CASCADE,
+        related_name="created_rh_comments",
+    )
+    updated_by = ForeignKey(
+        "user.User",
+        auto_created=True,
+        on_delete=CASCADE,
+        related_name="updated_rh_comments",
+    )
```
le-sun, line 21: Does `DefaultAuthenticatedModel` come with these fields?

src/researchhub_comment/related_models/rh_comment_thread_model.py
```diff
@@ -0,0 +1,23 @@
...
+        related_name="rh_threads",
+    )
+    thread_type = CharField(
+        max_length=144,
+        choices=RH_COMMENT_THREAD_TYPES,
+        default=GENERIC_COMMENT,
+    )
+    thread_reference = CharField(
+        blank=True,
+        help_text="""A thread may need a spcial referencing tool. Use this field for such a case""",
```
le-sun, line 20:
```suggestion
        help_text="""A thread may need a special referencing tool. Use this field for such a case""",
```

le-sun (changes requested):

src/researchhub_comment/related_models/rh_comment_model.py
calvinhlee23: We need to re-specify to form related_name for traversal reasons

PR #1203, opened 2023-01-19
[Comment Migration][3/n] - Backfill Signals

calvinhlee23:
Screen Shot 2023-01-19 at 3 29 19 PM


src/researchhub_comment/signals/rh_comment_signals.py
```diff
@@ -0,0 +1,186 @@
...
+        migrated_thread_comment.context_title = instance.context_title or ""
+        migrated_thread_comment.comment_content_src.save(
+            f"rh-comment-user-{upserter.id}-thread-{belonging_thread.id}-comment-{migrated_thread_comment.id}.txt",
+            comment_content_file
+        )
+        # intentionally querying DB to ensure that the instance was created properly
+        return RhCommentModel.objects.get(
+            legacy_id=legacy_thread_id,
+            legacy_model_type=LEGACY_THREAD,
+        )
```
le-sun, line 50: I think ```migrated_thread_comment.id``` works just as well. If the instance was not created properly, it wouldn't have a ID assigned to it. The same applies to the other signals down below.

le-sun (changes requested):

src/researchhub_comment/signals/rh_comment_signals.py
calvinhlee23: I also wanted the instances to be returned to for recursive call reasons. I would like to keep this.

PR #1207, opened 2023-02-14
Bounties in posts

yattias:
- Fixing query to support bounties in posts

FE: https://github.com/ResearchHub/researchhub-web/pull/1313

### DIFF
--- a/src/reputation/views/bounty_view.py
+++ b/src/reputation/views/bounty_view.py
@@ -54,21 +54,21 @@ class BountyViewSet(viewsets.ModelViewSet):
     ALLOWED_APPROVE_CONTENT_TYPES = ("thread", "comment", "reply")
 
     def _get_rh_fee_recipient(self):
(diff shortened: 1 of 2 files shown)


src/reputation/views/bounty_view.py
```diff
@@ -54,21 +54,21 @@ class BountyViewSet(viewsets.ModelViewSet):
     ALLOWED_APPROVE_CONTENT_TYPES = ("thread", "comment", "reply")
 
     def _get_rh_fee_recipient(self):
-        user = User.objects.filter(email="revenue@researchhub.com")
```
le-sun, line 57: These changes need to be reverted

le-sun (changes requested):

src/reputation/views/bounty_view.py
yattias: I gotta find a better solution to this. just a matter of time before I accidentally commit this to prod

PR #1211, opened 2023-02-21
Document Feed Filter

le-sun:
Filtering out papers that do not have hubs attached using the provided document filter


yattias:
@le-sun should we merge this or close this branch?

PR #1215, opened 2023-03-02
Celery Config Update

le-sun:
Reverting Celery timezone from PST to UTC


src/user/tasks.py
```diff
@@ -354,7 +352,7 @@ def notify_editor_inactivity():
 
 
 @periodic_task(
-    run_every=crontab(hour=15, minute=2),  # 3:02 PM PST (pst is system time)
+    run_every=crontab(hour=23, minute=5),  # 3:02 PM PST (pst is system time)
```
lightninglu10, line 355: Can you change the comment to be accurate, so we're not confused?

lightninglu10 (changes requested):
see comment

src/user/tasks.py
le-sun: Updated

PR #1218, opened 2023-03-09
[CommentMigration][2/n] Thread FilterSet, @Create Comment Actions &  Method Updates

calvinhlee23:
### Test: 
- go do http://localhost:8000/api/rh_comment_threads/

Screen Shot 2023-03-09 at 12 26 35 AM


src/hypothesis/views/citation_view.py
```diff
@@ -12,7 +13,7 @@
 from discussion.reaction_models import Vote as GrmVote
 
 
-class CitationViewSet(ModelViewSet, ReactionViewActionMixin):
+class CitationViewSet(ModelViewSet, ReactionViewActionMixin, RhCommentThreadViewMixin):
```
le-sun, line 16: I recently learned something new about Python inheritance: 
image

So we should correct the ordering of mixins and views.

src/hypothesis/views/hypothesis_view.py
```diff
@@ -17,7 +18,7 @@
 from utils.throttles import THROTTLE_CLASSES
 
 
-class HypothesisViewSet(ModelViewSet, ReactionViewActionMixin):
+class HypothesisViewSet(ModelViewSet, ReactionViewActionMixin, RhCommentThreadViewMixin):
```
le-sun, line 21: See previous review.

src/paper/views/paper_views.py
```diff
@@ -76,7 +77,7 @@
 from utils.throttles import THROTTLE_CLASSES
 
 
-class PaperViewSet(viewsets.ModelViewSet, ReactionViewActionMixin):
+class PaperViewSet(viewsets.ModelViewSet, ReactionViewActionMixin, RhCommentThreadViewMixin):
```
le-sun, line 80: See previous review.

src/researchhub_comment/related_models/rh_comment_model.py
```diff
@@ -77,23 +77,34 @@ def is_root_comment(self):
...
+        try:
+            rh_comment_serializer.is_valid(raise_exception=True)
+            rh_comment = rh_comment_serializer.save()
+            rh_comment.comment_content_src.save(
+                f"RH-THREAD-{rh_thread.id}-COMMENT-{rh_comment.id}-user-{data.user.id}.txt",
+                comment_content_src_file,
+            )
+            return rh_comment
+        except Exception as error:
+            raise Exception(f"Failed to RhCommentModel#create_from_data: {error}")
```
le-sun, line 107: We can remove this try catch, since it doesn't really do anything.

src/researchhub_comment/related_models/rh_comment_thread_model.py
```diff
@@ -35,18 +38,22 @@ class RhCommentThreadModel(AbstractGenericRelationModel):
...
+            return Citation
+        elif thread_target_model_name == "hypothesis":
+            return Hypothesis
+        elif thread_target_model_name == "paper":
             return Paper
-        if thread_content_model_name == "researchhub_post":
+        elif thread_target_model_name == "researchhub_post":
             return ResearchhubPost
         else:
+            import pdb; pdb.set_trace()
```
le-sun, line 56: pdb

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    citation_id = NumberFilter(
+        field_name="object_id",
+        label="citation_id",
+        method="get_qs_by_citation_id",
+    )
+
+    def get_qs_by_paper_id(self, qs, name, value):
+        return qs.filter(
+            content_type=get_content_type_for_model(Paper), object_id=int(value)
```
le-sun, line 83: No need to cast value as int

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def get_qs_by_researchhub_post_id(self, qs, name, value):
+        return qs.filter(
+            content_type=get_content_type_for_model(ResearchhubPost),
+            object_id=int(value),
```
le-sun, line 89: No need to cast value as int

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def get_qs_by_hypothesis_id(self, qs, name, value):
+        return qs.filter(
+            content_type=get_content_type_for_model(Hypothesis), object_id=int(value)
```
le-sun, line 94: No need to cast value as intNo need to cast value as int

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def get_qs_by_citation_id(self, qs, name, value):
+        return qs.filter(
+            content_type=get_content_type_for_model(Citation), object_id=int(value)
```
le-sun, line 99: No need to cast value as int

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    @property
+    def qs(self):
+        """
+        Intentionally hard limitting queryset to 10.
+        At current scale, it's hard to see that we need to fetch more than 10 threads all at once given a query.
+        This prevents malicious attempt to overload the server as well.
+        """
+        return super().qs[:10]
```
le-sun, line 109: This should be done in the view instead. Limiting in the pagination class.

src/researchhub_comment/views/rh_comment_thread_view.py
```diff
@@ -12,96 +13,29 @@
...
-            return (
-                valid_thread_target_model.object.get(id=thread_target_model_instance_id)
-                .rh_threads.filter(
-                    thread_reference=thread_reference, thread_type=thread_type
-                )
-                .first()
-            )
+    def get_filtered_queryset(self):
+        # NOTE: RhCommentThreadFilter has a qs limit of 10
+        return self.filter_queryset(self.get_queryset())
```
le-sun, line 41: See review above.

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -51,7 +52,7 @@
 from utils.sentry import log_error
 
 
-class ResearchhubPostViewSet(ModelViewSet, ReactionViewActionMixin):
+class ResearchhubPostViewSet(ModelViewSet, ReactionViewActionMixin, RhCommentThreadViewMixin):
```
le-sun, line 55: See review above about ordering.

src/researchhub_comment/related_models/rh_comment_model.py
```diff
@@ -77,23 +77,34 @@ def is_root_comment(self):
     """ --- METHODS --- """
 
     @classmethod
-    def create_from_request(cls, request, rh_thread):
-        request_data = request.data
+    def create_from_data(cls, data, rh_thread):
```
le-sun, line 80: Let's wrap this in an atomic transaction `with transaction.atomic():`, `from django.db import transaction`.

I wasn't really sure about having non retrieval db actions wrapped in an atomic block, but I think it's a good practice to do now, and it is something that I am trying to do now too.

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+    @action(detail=True, methods=["POST"], permission_classes=[AllowAny])
+    def get_comment_threads(self, request, pk=None):
+        raise NotImplementedError("Needs to be implemented")
+
+    def _retrieve_or_create_thread_from_request_data(self, request):
+        request_data = request.data
+        try:
+            thread_id = request_data.get("thread_id") or None
+            if thread_id is not None:
+                return RhCommentThreadModel.objects.get(id=int(thread_id))
```
le-sun, line 38: No need to cast thread_id

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+                    )
+        except Exception as error:
+            raise Exception(f"Failed to create / retrieve rh_thread: {error}")
+
+    def _get_existing_thread_from_request(self, request):
+        request_data = request.data
+        """ NOTE: sanity checking if payload included thread_id """
+        thread_id = request_data.get("thread_id") or None
+        if thread_id is not None:
+            return RhCommentThreadModel.objects.get(id=int(thread_id))
```
le-sun, line 64: No need to cast thread_id

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+                        RhCommentThreadModel.get_valid_thread_target_model(
+                            self._resolve_target_model_name(request)
+                        )
+                    )
+                    thread_target_instance = valid_thread_target_model.objects.get(
+                        id=self._resolve_target_model_instance_id(request)
+                    )
+                    return thread_target_instance.rh_threads.create(
+                        thread_type=request_data.get("thread_type"),
+                        thread_reference=request_data.get("thread_reference"),
```
le-sun, line 54: What does this do exactly?

le-sun (changes requested):

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
calvinhlee23: we do. The filter at form level casts the values into strings when converted to json

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
calvinhlee23: same here

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
calvinhlee23: some incoming form data (like postman? casts int to string. better to be safe

le-sun: I don't think it has any real benefit?

For example:
1. Int
```
In [22]: User.objects.get(id=2)
Out[22]: <User: 2: bank bank>
```
2. String
```
In [23]: User.objects.get(id="2")
Out[23]: <User: 2: bank bank>
```
3. String to Int
```
In [25]: User.objects.get(id=int("2"))
Out[25]: <User: 2: bank bank>
```
4. String to Int with error
```
In [26]: User.objects.get(id=int("\asfd\adsf"))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[26], line 1
----> 1 User.objects.get(id=int("\asfd\adsf"))

ValueError: invalid literal for int() with base 10: '\x07sfd\x07dsf'
```
5. String with error
```
In [27]: User.objects.get(id="\asfd\adsf")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
   ...
   ...
   2019 except (TypeError, ValueError) as e:
-> 2020     raise e.__class__(
   2021         "Field '%s' expected a number but got %r." % (self.name, value),
   2022     ) from e

ValueError: Field 'id' expected a number but got '\x07sfd\x07dsf'.
```

src/researchhub_comment/related_models/rh_comment_model.py
calvinhlee23: it does. It will show where the error happened in terms of code location & it will log the serializer validation stuff

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
calvinhlee23: There's no need to have a separate pagination class for threads at the moment. A document object should not have more than 5 threads idealy at the moment unless people start going crazy with inline commenting. I will add pagination there once we need to scale?

src/researchhub_comment/views/rh_comment_thread_view.py
calvinhlee23: Leaving a trace here to make sure that this doesn't get lost

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
calvinhlee23: Accepted. I'll try removing int casting for these places. But for now, I'm going to keep the int cast in the ViewFilter. It was error-ing without it

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
calvinhlee23: It grabs the target model class instance that is meant to create generic relation to Thread. (For example Paper class instance). Then creating a thread accordingly

src/researchhub_comment/related_models/rh_comment_model.py
le-sun: The view that's calling `create_from_data` already has a try catch

src/researchhub_comment/views/filters/rh_comment_thread_filters.py
le-sun: I think it's good practice to have the pagination class set if we want to hard limit the queryset. If a new programmer were to come on, it would be more logical for them to find it in the pagination class rather than somewhere that is relatively 'hidden' in the qs method.

src/researchhub_comment/related_models/rh_comment_model.py
```diff
@@ -77,29 +78,39 @@ def is_root_comment(self):
...
+        from researchhub_comment.serializers.rh_comment_serializer import (
+            RhCommentSerializer,
         )
-        rh_comment.comment_content_src.save(
-            f"RH-THREAD-{rh_thread.id}-COMMENT-{rh_comment.id}-user-{request.user.id}.txt",
-            comment_content_src_file,
-        )
-        rh_comment.refresh_from_db()
-        return rh_comment
+        with transaction.atomic():
```
le-sun, line 85: We can get rid of one of these atomic transactions. Either the one in the view or the mixin.

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
```diff
@@ -0,0 +1,128 @@
...
+from researchhub_comment.related_models.rh_comment_model import RhCommentModel
+from researchhub_comment.related_models.rh_comment_thread_model import (
+    RhCommentThreadModel,
+)
+
+
+class RhCommentThreadViewMixin:
+    @action(detail=True, methods=["POST"], permission_classes=[IsAuthenticated])
+    def create_rh_comment(self, request, pk=None):
+        with transaction.atomic:
```
le-sun, line 19:
```suggestion
        with transaction.atomic():
```

le-sun: See below review too

le-sun (changes requested):

src/researchhub_comment/related_models/rh_comment_model.py
calvinhlee23: Yes, but the point of having a class method is to be able to use it in more than one place. 
We cannot only rely on the call site.

le-sun: But is there a point of having it wrapped in a try block just to raise another exception?

src/researchhub_comment/views/rh_comment_thread_view_mixin.py
calvinhlee23: rm-ed

src/researchhub_comment/related_models/rh_comment_model.py
calvinhlee23: Yes, that way we know exactly where the exception was raised. From my experience, django doesn't bubble up the exceptions when it's caught at lower levels like this. This is good because it's won't double log & helps the devs to know exactly where it's throwing.

le-sun: I don't really see how this will help locate the error. In some cases, try catches usually make it harder to find where an exception is thrown (e.g., paper serializer/views ). I don't think this will be the case for this specific try catch, but if you see my example down below, catching the error makes no difference?

Another example:

Without a try/catch
```
  ...
  File "/Users/.../rh_comment_model.py", line 99, in create_from_data
    rh_comment_serializer.is_valid(raise_exception=True)
  ValidationError: 'foo'
```
vs

With a try/catch
```
  ...
  File "/Users/.../rh_comment_model.py", line 107, in create_from_data
    raise Exception(f"Failed to RhCommentModel#create_from_data: {error}")
  File "/Users/.../rh_comment_model.py", line 99, in create_from_data
    rh_comment_serializer.is_valid(raise_exception=True)
  Exception: 'Failed to RhCommentModel#create_from_data:  foo'
```

PR #1219, opened 2023-03-09
Function Based View - Popover

le-sun:
Updated popover endpoint to be a FBV

### DIFF
--- a/src/user/views/user_views.py
+++ b/src/user/views/user_views.py
@@ -824,7 +824,9 @@ def get_user_popover(request, pk=None):
                 "headline",
                 "profile_image",
             )
-        }
+        },
+        "usr_dus_get_editor_of": {"_include_fields": ("source",)},
+        "rag_dps_get_source": {"_include_fields": ("id", "name")},
     }
     serializer = DynamicUserSerializer(
         user,
@@ -832,6 +834,7 @@ def get_user_popover(request, pk=None):
         _include_fields=(
             "id",
             "author_profile",
+            "editor_of",
             "first_name",
             "last_name",
             "reputation",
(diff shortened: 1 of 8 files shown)


src/user/views/user_views.py
```diff
@@ -803,6 +804,45 @@ def sift_check_user_content(self, request):
             raise Exception("Sift verification signature mismatch")
 
 
+@api_view([RequestMethods.GET])
```
yattias, line 807: @le-sun will this speed up API requests 🤔 ?

yattias:
@le-sun looks great.
Is it possible to write two tests: 
- One that ensures that the endpoint returns data if ID is found
- One that ensures that the endpoint returns no dat if ID is not found?

Just as an assurance that these endpoints will alert us in case something breaks with the feature in the future

le-sun:
> @le-sun looks great. Is it possible to write two tests:
> 
> * One that ensures that the endpoint returns data if ID is found
> * One that ensures that the endpoint returns no dat if ID is not found?
> 
> Just as an assurance that these endpoints will alert us in case something breaks with the feature in the future

Added 2 tests.
1. First test checks if the status code == 200 and contains user data
2. Second test checks if the endpoint will 404

src/user/views/user_views.py
le-sun: To sum up what we went over during the meeting:

Because class based views has a lot of inheritance, it adds some overhead when processing a request. Function based views are relatively barebones, so there is a slight performance boost.

CBVs are nice to have because of inheritance. It provides a lot of default methods that we don't have to implement ourselves when creating a view.
FBVs are simple, which means it should be used for simple endpoints. Adding more complex logic in FBVs is redundant since CBVs provide a lot of utility for complex logic.

PR #1221, opened 2023-03-17
Adding rsc exchange rate info

lightninglu10:
endpoint to return exchange rate

### DIFF
--- a/src/purchase/serializers.py
+++ b/src/purchase/serializers.py
@@ -19,7 +19,14 @@ from discussion.serializers import (
     ThreadSerializer,
 )
 from paper.serializers import BasePaperSerializer, DynamicPaperSerializer
-from purchase.models import AggregatePurchase, Balance, Purchase, Support, Wallet
+from purchase.models import (
+    AggregatePurchase,
+    Balance,
+    Purchase,
+    RscExchangeRate,
+    Support,
+    Wallet,
+)
 from reputation.models import Bounty, Distribution, Withdrawal
 from reputation.serializers import (
     BountySerializer,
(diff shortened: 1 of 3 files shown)


src/purchase/serializers.py
```diff
@@ -394,3 +401,14 @@ def get_stats(self, purchase):
...
+
+class RscExchangeRateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = RscExchangeRate
+        fields = "__all__"
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        data["created_date"] = instance.created_date.isoformat()
+        return data
```
le-sun, line 414: Is this necessary?

le-sun: 1. Without overriding the `to_representation` method ```"created_date": "2023-02-24T19:43:27.098203Z"```
2. Overriding the method ```"created_date": "2023-02-24T19:43:27.098203+00:00"```

le-sun: If that specific format is necessary, can we do this instead?

```
class RscExchangeRateSerializer(serializers.ModelSerializer):
    created_date = serializers.DateTimeField(format="%Y-%m-%dT%H:%M:%S.%f%z")

    class Meta:
        model = RscExchangeRate
        fields = "__all__"

```

This will return ```"created_date": "2023-02-24T19:43:27.098203+0000"```

le-sun (changes requested):

src/purchase/serializers.py
lightninglu10: I'm OK removing. will do that then merge

PR #1229, opened 2023-04-09
PDF metadata and download improvements

yangzh:
This piece of work add supports for web-scraping certain web pages to extract metadata for academic literature, as well as download them (PDF) from certain aggregation sites (after the download url is extracted).


yangzh:
OK, @yattias The whole work is done, including code, documentation, test (both static and live scraping, as agreed). Please take a look.

PR #1279, opened 2023-05-15
[RM][Usability] - Permissions on Reference Manager

calvinhlee23:
see title


src/citation/views/citation_project_view.py
```diff
@@ -1,25 +1,36 @@
...
 
 class CitationProjectViewSet(ModelViewSet):
     queryset = CitationProject.objects.all()
     permission_classes = [IsAuthenticated]
     serializer_class = CitationProjectSerializer
-    ordering = ["-updated_date"]
+    ordering = ["created_date"]
+
+    # NOTE: use upsert instead
+    def create(self, _request):
```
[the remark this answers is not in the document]
calvinhlee23: yes. Otherwise, it uses default serializer methods.

src/citation/views/citation_project_view.py
```diff
@@ -49,3 +64,28 @@ def list(self, request):
                 error,
                 status=status.HTTP_400_BAD_REQUEST,
             )
+
+    @action(detail=True, methods=["POST"], permission_classes=[IsAuthenticated])
+    def upsert(self, request, pk=None, *args, **kwargs):
+        # with transaction.atomic:
```
le-sun, line 70: Let's bring back the atomicity.

src/citation/related_models/citation_project.py
```diff
@@ -33,12 +38,45 @@ class CitationProject(DefaultAuthenticatedModel):
...
+                editor_exists = self.permissions.filter(
+                    access_type=EDITOR, user=editor_id
+                ).exists()
+                if not editor_exists:
+                    self.permissions.create(
+                        access_type=EDITOR, user=User.objects.get(id=editor_id)
+                    )
+            return True
+        except Exception as error:
+            return False
```
le-sun, line 55: What's the point of capturing the errors? We should log these too with sentry.

src/citation/serializers/citation_project_serializer.py
```diff
@@ -23,16 +28,39 @@ class Meta:
...
+            User.objects.filter(id__in=editor_ids), many=True
+        ).data
+
     def get_current_user_has_access(self, project_instance):
         try:
             current_user = self.context.get("request").user
             return project_instance.get_current_user_has_access(current_user)
-        except Exception as error:
+        except Exception as _error:
             pass
```
le-sun, line 52: Why the try/catch? Also, return false?

src/citation/related_models/citation_project.py
```diff
@@ -33,12 +38,45 @@ class CitationProject(DefaultAuthenticatedModel):
...
             return True
         else:
             return self.permissions.filter(user=user).exists()
+
+    def set_creator_as_admin(self):
+        try:
+            self.permissions.create(access_type=ADMIN, user=self.created_by)
+            return True
+        except Exception as error:
+            return False
```
le-sun, line 82: Same as above

src/citation/views/citation_entry_view.py
```diff
@@ -29,14 +30,18 @@ def user_citations(self, request):
...
         project_id = request.GET.get("project_id")
+        citations_query = None
         if project_id:
             citations_query = CitationProject.objects.get(id=project_id).citations.all()
         elif organization_id:
             citations_query = Organization.objects.get(
                 id=organization_id
             ).created_citations.all()
+            if request.query_params.get("get_current_user_citations", None):
+                citations_query = citations_query.filter(created_by=user.id)
```
le-sun, line 41: We should do filtering with a `filter_class` instead. See `researchhub_comment/filters.py` for a recent example.

src/citation/views/citation_entry_view.py
```diff
@@ -29,14 +30,18 @@ def user_citations(self, request):
...
         else:
-            citations_query = user.created_citation_citationentry.all()
+            raise PermissionError("Fetch not allowed without org_id or project_id")
```
le-sun, line 43: If we add a `filter_class`, we can set the field to be required and we won't need to manually raise this exception.

le-sun (changes requested):

le-sun (approved):

PR #1285, opened 2023-05-26
dedupe pdf upload citations

lightninglu10:


src/citation/serializers/citation_entry_serializer.py
```diff
@@ -29,6 +29,7 @@ class Meta:
     """ ----- Django Method Overrides -----"""
 
     def create(self, validated_data):
+        CitationEntry.objects.filter(doi=validated_data.get("doi")).exists()
```
le-sun, line 32: Hanging line of code?

PR #1291, opened 2023-06-09
Document Metadata

le-sun:
Created endpoint to return "live" document meta data
Removed unused serializer fields from existing serializers


src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -655,6 +655,73 @@ def check_user_vote(self, request):
                     ).data
         return Response(response, status=status.HTTP_200_OK)
 
+    def _get_document_metadata_context(self):
```
yattias, line 658: 💯 I like the declarative nature of this function

yattias (approved):

PR #1295, opened 2023-06-13
[Ref Manager] Citation List Privacy Update 

calvinhlee23:
- Org Refs now pull entries that belong to orgs directly or that belong to projects that are public. 
-  My Refs now pull all created entries by the current user

### DIFF
--- a/src/citation/views/citation_entry_view.py
+++ b/src/citation/views/citation_entry_view.py
@@ -1,11 +1,13 @@
-from django.db import transaction
 from django_filters.rest_framework import DjangoFilterBackend
+from django.db import transaction
+from django.db.models import Q
 from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.filters import OrderingFilter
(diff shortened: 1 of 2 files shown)


src/citation/views/citation_entry_view.py
```diff
@@ -68,11 +70,14 @@ def user_citations(self, request):
...
-            ).created_citations.all()
             if request.query_params.get("get_current_user_citations", None):
-                citations_query = citations_query.filter(created_by=user.id)
+                citations_query = citations_query.filter(
+                    organization=organization_id, created_by=user.id
+                )
+            else:
+                citations_query = Organization.objects.get(
+                    id=organization_id
+                ).created_citations.filter(Q(project__is_public=True) | Q(project=None))
```
le-sun, line 80: We have the filters class for citations, where we should move this logic to

le-sun (changes requested):

le-sun (changes requested):

le-sun (approved):

PR #1296, opened 2023-06-14
Citation Manager: Adding in link to paper 

lightninglu10:
- Adding in a link to the paper per citation
- Also returning a serialized version of a project parent

### DIFF
--- a/src/citation/serializers/citation_entry_serializer.py
+++ b/src/citation/serializers/citation_entry_serializer.py
@@ -22,7 +22,7 @@ class CitationEntrySerializer(DefaultAuthenticatedSerializer):
     checksum = ReadOnlyField()
     fields = JSONField()
     required_fields = SerializerMethodField(read_only=True)
-    paper = PaperSerializer()
+    paper = SerializerMethodField(read_only=True)
 
     class Meta:
         model = CitationEntry
@@ -66,6 +66,10 @@ class CitationEntrySerializer(DefaultAuthenticatedSerializer):
         return fields_data
 
     """ ----- Serializer Methods -----"""
+
+    def get_paper(self, citation_entry):
+        if citation_entry.unified_doc:
+            return PaperSerializer(citation_entry.unified_doc.paper)
 
     def get_attachment_url(self, citation_entry):
         try:
--- a/src/citation/views/citation_project_view.py
+++ b/src/citation/views/citation_project_view.py
@@ -43,7 +43,7 @@ class CitationProjectViewSet(ModelViewSet):
         )
 
     def update(self, request, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


src/citation/related_models/citation_entry_model.py
```diff
@@ -26,6 +27,13 @@ class CitationEntry(DefaultAuthenticatedModel):
     organization = models.ForeignKey(
         Organization, related_name="created_citations", on_delete=models.CASCADE
     )
+    paper = models.ForeignKey(
```
calvinhlee23, line 30: Wonder if we should just link it `rh_uni_doc` instead. 
In case we decide to allow users to cite researchhub contents in the future, it's better to get ahead of this and allow unified document instead of trying to add individual document types.

calvinhlee23 (changes requested):
Requesting change for visibility.

src/citation/utils.py
```diff
@@ -50,7 +52,12 @@ def create_paper_from_citation(citation):
     process_id = None
 
     if not duplicate_papers:
-        data = {"uploaded_by": citation.created_by.id, "url": url}
+        data = {
+            # "uploaded_by": citation.created_by.id,
+            "uploaded_by": None,
```
le-sun, line 57: Is there a reason to leave `uploaded_by` as None?

src/citation/related_models/citation_entry_model.py
le-sun: \+ 1

le-sun (changes requested):

src/citation/serializers/citation_entry_serializer.py
```diff
@@ -21,6 +22,7 @@ class CitationEntrySerializer(DefaultAuthenticatedSerializer):
     checksum = ReadOnlyField()
     fields = JSONField()
     required_fields = SerializerMethodField(read_only=True)
+    paper = PaperSerializer()
```
calvinhlee23, line 25: oh :o how does this work? The model has UniDoc. Does this automagically infer?

src/citation/serializers/citation_project_serializer.py
```diff
@@ -10,8 +8,15 @@
 from utils.serializers import DefaultAuthenticatedSerializer
 
 
+class ParentSerializer(DefaultAuthenticatedSerializer):
```
calvinhlee23, line 11: nice

calvinhlee23 (approved):
Thanks for updating the PR & adding unidoc to citation_entry. 
Can you take a look at my comment on the citation_entry_serializer?
Overall, looks good.

le-sun (approved):

PR #1297, opened 2023-06-14
[Citation Manager] PDF Uploads

le-sun:
Integrating with Grobid service
Generating pre-signed url for pdf uploads
Added callback endpoint to hook in with S3 Lambda trigger


src/citation/exceptions.py
```diff
@@ -0,0 +1,14 @@
+from utils.exceptions import Error
+
+
+class GrobidProcessingError(Error):
```
yattias, line 4: 💯  for specific error

src/citation/permissions.py
```diff
@@ -15,3 +16,13 @@ def has_permission(self, request, view):
...
         requestor = request.user
         return obj.get_is_user_admin(requestor)
+
+
+class PDFUploadsS3CallBack(BasePermission):
+    message = "Permission Denied: Endpoint restricted to S3 Lambda trigger"
+
+    def has_permission(self, request, view):
+        client_ip = get_client_ip(request)
+        if client_ip not in LAMBDA_IPS:
```
yattias, line 26: Good thinking, I imagine we were able to get a static list of ips?

src/citation/views/citation_entry_view.py
```diff
@@ -44,22 +44,60 @@ def retrieve(self, request):
 
     @action(detail=False, methods=["post"], permission_classes=[IsAuthenticated])
     def pdf_uploads(self, request):
```
yattias, line 46: Minor comment: Can we change to `upload_pdf`, `upload_files` or `upload_pdfs` ? I think having a verb in the name of the action would be better

yattias (approved):

src/citation/views/citation_entry_view.py
le-sun: Will do

src/citation/permissions.py
le-sun: Yeah, I configured the Lambda triggers to be connected within a couple of subnets in our VPC. Each subnet is connected to NAT gateway and each gateway has an elastic ip addressed attached to it. There are 2 elastic addresses assigned, one for prod and the other for staging.

src/citation/schema.py
```diff
@@ -13,12 +13,11 @@
...
 
-def generate_json_for_journal(pdf2doi):
-    schema = generate_schema_for_citation(JOURNAL_ARTICLE)
+def generate_json_for_doi(doi):
     json = {}
-    doi_string = pdf2doi["identifier"]
+    schema = generate_schema_for_citation(JOURNAL_ARTICLE)
     open_alex = OpenAlex()
-    result = open_alex.get_data_from_doi(doi_string)
+    result = open_alex.get_data_from_doi(doi)
```
calvinhlee23, line 20: how do we assume they are journal article?

src/citation/schema.py
```diff
@@ -13,12 +13,11 @@
...
     for field in schema["required"]:
         mapping_field = OPENALEX_JOURNAL_MAPPING.get(field, "")
```
calvinhlee23, line 22: OPENALEX_JOURNAL_MAPPING. Where did we get this from?

src/citation/utils.py
```diff
@@ -1,34 +1,87 @@
...
+        "bucket_name": f"{AWS_STORAGE_BUCKET_NAME}",
+    }
+    try:
+        response = requests.post(url, data=request_data, timeout=10)
+        data = response.json()
+        status = data.get("status")
+        if status == 200:
+            return data.get("data", {})
+    except requests.ConnectionError as e:
+        log_error(e)
```
calvinhlee23, line 36: could you add a suffix or beginning - phrases here so that we can easily identify what the error was for when we look in sentry without reading through the errors at first glance?

src/citation/views/citation_entry_view.py
```diff
@@ -43,23 +43,61 @@ def retrieve(self, request):
...
+                    "x-amz-meta-project-id": f"{project_id}",
+                    "x-amz-meta-file-name": filename,
+                },
+            },
+            ExpiresIn=60 * 5,
+        )
+        return Response(res, status=200)
+
+    @action(detail=False, methods=["post"], permission_classes=[PDFUploadsS3CallBack])
+    def upload_pdfs_callback(self, request):
```
calvinhlee23, line 84: is there a way to secure this endpoing so that the we only allow request from recognized sources? in this case S3

calvinhlee23: Not sure if this is fail safe or how much logging we do at this level. But I just want to avoid this endpoint getting pinged from random sources

calvinhlee23 (changes requested):
Requesting changes for visibility. 
Architecture is very sound. I just had some questions

src/citation/schema.py
le-sun: A lot of this code is originally from https://github.com/ResearchHub/researchhub-backend/pull/1283/files

src/citation/utils.py
le-sun: I'll make that change

src/citation/views/citation_entry_view.py
le-sun: This is a good observation. I have a permission class set to only allow incoming requests from our S3 trigger, so it can't be pinged from random sources (HTTP 403).

src/citation/utils.py
```diff
@@ -1,34 +1,87 @@
...
+        raise GrobidProcessingError(e, "Request to Grobid server timed out")
+
+    raise GrobidProcessingError(
+        "Could not extract data", "Grobid queue is most likely full"
+    )
+
+
+def get_citation_entry_from_pdf(path, filename, user_id, organization_id, project_id):
```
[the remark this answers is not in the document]
le-sun: Yes. We also make the assumption that a uploaded pdf is a journal article. I'm not sure if this is a safe assumption, but that's how the code was like before.

src/citation/schema.py
le-sun: Made a comment further down.

PR #1299, opened 2023-06-16
[Reviews] Adding content_type, object_id to reviews

yattias:
- Adding `content_type`, `object_id` to `reviews` table
- Adding `review` to comment serializer


src/review/views/review_view.py
```diff
@@ -34,6 +35,10 @@ def create(self, request, *args, **kwargs):
         unified_document = ResearchhubUnifiedDocument.objects.get(id=args[0])
         request.data["created_by"] = request.user.id
         request.data["unified_document"] = unified_document.id
+        request.data["content_type"] = ContentType.objects.get(
+            model=request.data.get("content_type", None)
```
le-sun, line 39: We should have an allowed list of accepted content types, so a user won't be able to create a review of another model. The bounty view set (line 179) should have an example.

src/researchhub_comment/serializers/rh_comment_serializer.py
```diff
@@ -164,3 +166,15 @@ def get_user_vote(self, comment):
             return vote
         except Vote.DoesNotExist:
             return None
+
+        return None
+
+    def get_review(self, comment):
+        context = self.context
+        _context_fields = context.get("rhc_dcs_get_review", {})
+        review = comment.unified_document.reviews.filter(object_id=comment.id).first()
```
le-sun, line 175: If we create a generic relation on the comment model, we can simplify this to something like
```suggestion
        review = comment.reviews.first()
```
and the model class will look something like
```
reviews = GenericRelation("review.Review")
```
You can also see an example in the bounty model class (line 60-61)

le-sun (changes requested):

src/review/views/review_view.py
```diff
@@ -29,11 +31,19 @@ class ReviewViewSet(viewsets.ModelViewSet, ReactionViewActionMixin):
...
     queryset = Review.objects.all()
     ordering = ("-created_date",)
+    ALLOWED_CONTENT_TYPES = "rhcommentmodel"
 
     def create(self, request, *args, **kwargs):
         unified_document = ResearchhubUnifiedDocument.objects.get(id=args[0])
         request.data["created_by"] = request.user.id
         request.data["unified_document"] = unified_document.id
+
+        if request.data.get("content_type") not in ReviewViewSet.ALLOWED_CONTENT_TYPES:
```
le-sun, line 41:
```suggestion
        if request.data.get("content_type") not in self.ALLOWED_CONTENT_TYPES:
```
Also `ALLOWED_CONTENT_TYPES` needs to be in a list

le-sun (changes requested):
Small change

le-sun (approved):

PR #1304, opened 2023-06-23
Support for inline comments

yattias:
- Adding `anchor` attribute that will store serialized position of the anchor (e.g. highlight) of the comment
- Filtering by inline comments


le-sun (approved):
Looks good. Are we keeping in `thread_reference` for now?

PR #1309, opened 2023-06-30
QOL for Development with Citation Manager 

le-sun:
Removed exception for uploading pdfs in a dev environment
Added localhost to allowed lambda ips
Fixed bug with non compliant ascii filenames


calvinhlee23 (approved):
looks good to me

PR #1325, opened 2023-07-18
create initial models, migrations, serializers and endpoints for HubV2

florin-chelaru:
Introduces the following endpoints (via `HubV2ViewSet`):
`GET /api/hubs/` (list hubs + pagination)
`GET /api/hubs/<hub_id>` (get hub by id)

Write endpoints are disallowed.

Models: 
* `HubProvider`
* `HubV2`
* `HubMetadata`


src/hub/tests/test_views.py
```diff
@@ -208,3 +207,186 @@ def get_invite_to_hub_response(self, user, hub, emails):
             data,
             headers={'HTTP_ORIGIN': 'researchhub.com'}
         )
+
+
+class HubV2ViewsTests(APITestCase):
```
yattias, line 212: Thanks for adding tests!

florin-chelaru:
@yattias answers:
1. No. The existing endpoints are actually `/api/hub/` (singular). The new implementation adds `/api/hubs/`. I want to slowly transition to everything plural, to match the best practices described here: https://google.aip.dev/122#collection-identifiers (use plurals for collection identifiers). In the case of hubs, this is a good opportunity to make the transition more seamlessly, without having to expose any "V2" in the API interface.
2. Both endpoint sets are available now: `/api/hub/` gets the old hubs, and `/api/hubs/` the new ones.

src/pyproject.toml
```diff
@@ -78,6 +78,7 @@ django-elasticsearch-dsl = "7.2.2"
 django-elasticsearch-dsl-drf = "0.20.8"
 q5-django-inlinecss = "0.0.2"
 pdf2doi = "1.5"
+django-clearcache = "^1.2.1"
```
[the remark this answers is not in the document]
yattias: Gotcha. Yeah that sounds reasonable. Let's keep it. Curious though if we already have a command line mechanism @le-sun ?

src/hub/models.py
```diff
@@ -138,3 +136,38 @@ class HubMembership(models.Model):
 
     created_date = models.DateTimeField(auto_now_add=True)
     updated_date = models.DateTimeField(auto_now=True)
+
+
+# HubV2
```
le-sun, line 141: We've been moving away from keeping all our models within `models.py`. We don't have a codified system, but if you check out some of our latest Django applications (researchhub_comment, citation_entry), you can see how we created a folder named `related_models` and created separate files for each model.

le-sun (changes requested):
One thing that I realized is that our setup documentation doesn't have our auto linting/formatting installation. If you run `pre-commit install`, that should install our formatting tools.

florin-chelaru:
> One thing that I realized is that our setup documentation doesn't have our auto linting/formatting installation. If you run pre-commit install, that should install our formatting tools.

It doesn't seem to be working, or doesn't have enough rules: imports are not automatically optimized (sorted and unused imports removed) and lines are not truncated. Currently I'm using flake8 to take care of that, but it's not done automatically.

florin-chelaru:
> > One thing that I realized is that our setup documentation doesn't have our auto linting/formatting installation. If you run pre-commit install, that should install our formatting tools.
> 
> It doesn't seem to be working, or doesn't have enough rules: imports are not automatically optimized (sorted and unused imports removed) and lines are not truncated. Currently I'm using flake8 to take care of that, but it's not done automatically.

Actually I take that back. It did work, and formatting was done. Thanks for the heads up @le-sun !

src/hub/models.py
florin-chelaru: Done

src/hub/views.py
```diff
@@ -466,3 +458,30 @@ class HubCategoryViewSet(viewsets.ModelViewSet):
...
     permission_classes = [IsAuthenticatedOrReadOnly]
+
+
+class HubV2Pagination(pagination.PageNumberPagination):
+    page_size = 10
+    page_size_query_param = 'page_size'
+    max_page_size = 10000
+
+
+class HubV2ViewSet(viewsets.ModelViewSet):
```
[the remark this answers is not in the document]
florin-chelaru: Done

src/hub/models.py
```diff
@@ -138,3 +136,38 @@ class HubMembership(models.Model):
...
+    updated_date = models.DateTimeField(auto_now=True)
+    is_removed = models.BooleanField(default=False)
+
+    def __str__(self):
+        return self.display_name
+
+
+class HubMetadata(models.Model):
+    hub = models.ForeignKey(HubV2, on_delete=models.CASCADE)
+    hub_provider = models.ForeignKey(HubProvider, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
florin-chelaru: I think I got it. Added `related_name` for both fields.

yattias (approved):
This looks good to me

src/hub/related_models/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .hub_metadata import HubMetadata
+from .hub_provider import HubProvider
+from .hub_v2 import HubV2
```
le-sun, line 3: These imports need to be in `models.py` as well.

I know this is a bit out of scope for the project, but if you could separate out the current models in `models.py` into their respective files, that would be great too.

le-sun (changes requested):
Minor change

PR #1326, opened 2023-07-19
Increasing comment page size to support inline comments

yattias:
Large max page size needed for inline comments since they are not paginated and need to be load all at once


le-sun (approved):
Will this affect anything else? Like what the frontend uses right now to fetch comments?

yattias:
@le-sun the frontend currently controls the pagination in different context. For normal conversation stuff, I am passing a small page size of 20. For inline comments, I'd like that page size to be much higher. TLDR; won't affect existing functionality

PR #1336, opened 2023-08-03
Minor Bugfixes

le-sun:
Added inline comments to document metadata filtering
Added thread anchor to live feed serializer context


yattias (approved):
Looks good

PR #1351, opened 2023-08-29
Slugify folders

lightninglu10:
Slug to folders


src/citation/signals.py
```diff
@@ -0,0 +1,6 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from django.utils.crypto import get_random_string
+from django.utils.text import slugify
+
+from .models import CitationProject
```
le-sun, line 6: We can get rid of this file since we are slugifying on create.

src/citation/views/citation_project_view.py
```diff
@@ -32,6 +34,19 @@ def create(self, request, *args, **kwargs):
...
             project.add_editors(upserted_collaborators.get("editors", []))
             project.add_viewers(upserted_collaborators.get("viewers", []))
+            suffix = get_random_string(length=32)
+            slug = slugify(project.project_name)
+            if not slug:
+                slug += suffix
+            if not CitationProject.objects.filter(slug=slug).exists():
+                project.slug = slug
+            else:
+                project.slug = slug + "-" + suffix
```
le-sun, line 44:
```suggestion
                project.slug = f"{slug}_{suffix}"
```

le-sun (changes requested):
Few comments

PR #1355, opened 2023-09-08
Hubs V2

yattias:
- Update concept model and created the proper relationship to concept data to hubs
- Backfill script to fetch concepts, create associated hubs and relationship between hub <> concept
- Created hub suggester for elastic search to be used as autocomplete on FE


src/paper/management/commands/backfill_concepts.py
```diff
@@ -0,0 +1,76 @@
...
+            papers = Paper.objects.filter(created_date__gte=start_date)
+            print(f"Found {papers.count()} papers created after {start_date_str}")
+        elif doi:
+            papers = Paper.objects.filter(doi=doi)
+        else:
+            self.stdout.write(self.style.ERROR("Please provide a start date or DOI."))
+            return
+
+        for paper in papers:
+            try:
```
lightninglu10, line 35: Can we also print here how many are left to do? I think this will probably take a long time, so knowing how many entries are left would be good

src/utils/openalex.py
```diff
@@ -44,51 +44,38 @@ def get_data_from_doi(self, doi):
             raise DOINotFoundError(f"No OpenAlex works found for doi: {doi}")
         return results[0]
 
-    def get_data_from_source(self, source, date, page=1):
```
lightninglu10, line 47: removing this caused an error here: https://quantfive.sentry.io/issues/4464676335/?referrer=alert_email&alert_type=email&alert_timestamp=1694206809609&alert_rule_id=885988&notification_uuid=15d9afb5-c33c-4bb1-a19a-1869125cccfc&environment=staging

src/hub/models.py
```diff
@@ -72,13 +65,18 @@ class Hub(models.Model):
     paper_count = models.IntegerField(default=0)
     discussion_count = models.IntegerField(default=0)
 
-    is_removed = models.BooleanField(
-        default=False,
-        help_text=HELP_TEXT_IS_REMOVED
+    is_removed = models.BooleanField(default=False, help_text=HELP_TEXT_IS_REMOVED)
+
+    concept = models.OneToOneField(
```
lightninglu10, line 70: Been thinking about this a little bit, and I wonder if we should just merge these two? Like not even have the Concept model and instead put everything from the concept model into Hub.

My rationale is it makes our code less complex to have just one model representing this idea, rather than 2. 

@le-sun would like your thoughts on this too.

src/utils/openalex.py
yattias: I'll check this out, might have removed it as part of the rebase

yattias: I think it is good to keep these separate, at least for the time being. Concept and hub have overlap but are different enough where it makes sense to segregate the two.

For reference, we have the concept table

image

And the unified doc concept table created through the oneToOne relationship
image

src/hub/models.py
yattias: I think it is good to keep these separate, at least for the time being. Concept and hub have overlap but are different enough where it makes sense to segregate the two.

For reference, we have the concept table

image

And the unified doc concept table created through the oneToOne relationship
image

lightninglu10: The part that's slightly confusing to me is having a separate unified doc concept. 

Now we have 3 places "hubs" lives and it's not evident to me what dictates our queries

yattias: Yeah, I get your point. Three seems like a lot. 

I'll explain:

We need the unified doc association because it has **specific** attributes for that paper - such as `relevancy_score` and `level` These don't belong on either concept or hub because they are specific to the paper.

It is defined as such on `UnifiedDocumentModel`

```
    concepts = models.ManyToManyField(
        "tag.Concept",
        related_name="documents",
        blank=True,
        through="UnifiedDocumentConcepts",
    )
```

As for the concept table. It has a bunch of specific attributes to OpenAlex. I'd like to keep them separate so that their world doesn't fully influence our world - at least not quite yet but I do think there is an opportunity in the future to merge.

The good news is that the frontend will not be aware of these concepts. The FE will only be dealing with hubs.

src/hub/serializers.py
```diff
@@ -164,3 +165,22 @@ def get_editor_permission_groups(self, hub_instance):
...
             many=True,
         ).data
+
+    def get_relevancy_score(self, hub_instance):
+        from researchhub_document.models import UnifiedDocumentConcepts
+
+        unified_document = self.context.get("unified_document", None)
+        relevancy_score = 0
+
+        if unified_document and hub_instance.concept:
```
le-sun, line 175:
```python
if unified_document and concept := getattr(hub_instance, "concept", None):

```

le-sun: I think `hub_instance.concept` will throw an error if it happens to be null. This is because it's an 1-1 relationship, compared to a foreign key, which would usually return `None`.

src/hub/serializers.py
```diff
@@ -164,3 +165,22 @@ def get_editor_permission_groups(self, hub_instance):
...
+            associated_concept = unified_document.concepts.filter(
+                concept=hub_instance.concept.id
+            ).first()
+
+            concept_with_relevancy = UnifiedDocumentConcepts.objects.filter(
+                unified_document=unified_document, concept=associated_concept
+            ).first()
```
le-sun, line 182:
```python
concept_with_relevancy = UnifiedDocumentConcepts.objects.filter(unified_document=unified_document, concept=concept.id).first()
```

le-sun (changes requested):

src/paper/paper_upload_tasks.py
```diff
@@ -454,14 +455,14 @@ def celery_openalex(self, celery_data):
                 "pdf_license_url": url,
                 "external_source": host_venue.get("display_name", None),
             }
+
+            print("raw_authors", raw_authors)
```
lightninglu10, line 459: remove print here

src/hub/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+from hub.models import Hub
+from researchhub.celery import QUEUE_HUBS
+
+
+@periodic_task(
+    run_every=crontab(minute=0, hour=0),
+    priority=5,
+    queue=QUEUE_HUBS,
+)
+def calculate_and_set_hub_counts():
```
le-sun, line 17: I had an idea when looking at this script. I'm not sure if this works, but it can be a potential optimization:

```python
    hubs = Hubs.objects.annotate(
        total_paper_count=Count("related_documents"),
        posts_discussion_count=Sum("related_documents__posts__discussion_count"),
        paper_discussion_count=Sum("related_documents__paper__discussion_count"),
    ).update(
        paper_count=F("total_paper_count"),
        discussion_count=F("posts_discussion_count") + F("paper_discussion_count")
    )
```

This might allow us to skip over the iterator and have everything done in SQL

src/paper/paper_upload_tasks.py
le-sun, line 459: Delete

src/paper/paper_upload_tasks.py
```diff
@@ -608,13 +586,11 @@ def celery_create_paper(self, celery_data):
                 f"Unable to find article for: {dois}, {paper_submission.url}"
             )
 
-        concepts = paper_data.pop("concepts", None)
-
         async_paper_updator = getattr(paper_submission, "async_updator", None)
         paper = Paper(**paper_data)
         if async_paper_updator is not None:
             paper.doi = async_paper_updator.doi
-            paper.hub.add(*async_paper_updator.hubs)
```
le-sun, line 617: Maybe we should keep this for backwards compatibility?

src/paper/paper_upload_tasks.py
```diff
@@ -677,12 +633,32 @@ def celery_create_paper(self, celery_data):
...
+                through_defaults={
+                    "relevancy_score": paper_concept["score"],
+                    "level": paper_concept["level"],
+                },
+            )
+            hubs = Hub.objects.filter(concept=concept)
+            paper.unified_document.hubs.add(*hubs)
+
+    except Exception as e:
+        print("Failed to save concepts for paper" + str(paper.id))
```
le-sun, line 657: We should log this in sentry

src/paper/paper_upload_tasks.py
```diff
@@ -677,12 +633,32 @@ def celery_create_paper(self, celery_data):
...
+            concept = Concept.create_or_update(paper_concept)
+            paper.unified_document.concepts.add(
+                concept,
+                through_defaults={
+                    "relevancy_score": paper_concept["score"],
+                    "level": paper_concept["level"],
+                },
+            )
+            hubs = Hub.objects.filter(concept=concept)
+            paper.unified_document.hubs.add(*hubs)
```
le-sun, line 654:
```suggestion
            paper.unified_document.hubs.add(concept.hub)
```

src/paper/serializers/paper_serializers.py
```diff
@@ -442,9 +442,9 @@ def create(self, validated_data):
 
                 self._add_orcid_authors(paper)
                 paper.hubs.add(*hubs)
-                for hub in hubs:
-                    hub.paper_count = hub.get_paper_count()
-                    hub.save(update_fields=["paper_count"])
+                # for hub in hubs:
+                #     hub.paper_count = hub.get_paper_count()
+                #     hub.save(update_fields=["paper_count"])
```
le-sun, line 447: We can just delete this whole block of code

src/paper/serializers/paper_serializers.py
```diff
@@ -545,12 +545,6 @@ def update(self, instance, validated_data):
                     paper.hubs.add(*hubs)
                     unified_doc.hubs.remove(*remove_hubs)
                     unified_doc.hubs.add(*hubs)
-                    for hub in remove_hubs:
-                        hub.paper_count = hub.get_paper_count()
-                        hub.save(update_fields=["paper_count"])
-                    for hub in new_hubs:
-                        hub.paper_count = hub.get_paper_count()
-                        hub.save(update_fields=["paper_count"])
```
le-sun, line 553: We can delete this too

le-sun (changes requested):

src/hub/serializers.py
yattias: Seems like my version of python doesn't have the walrus operator even though I'm in 3.8. But, I will make the change (without the walrus)

src/hub/tasks.py
yattias: It's a good idea. Seems to not work for me. I'll have to debug this further

le-sun (approved):
Looks good. There is a merge conflict with a migration file we need to solve first.

PR #1357, opened 2023-09-13
PDF Compression and Linearization with S3 -> Cloudfront Distribution

le-sun:
Calling Lambda Ghostscript function on PDFs and user citations to compress and linearize them
Using Cloudfront CDN for file delivery


lightninglu10:
Is the lambda script in this PR? I think we should have it in version control

le-sun:
> Is the lambda script in this PR? I think we should have it in version control

https://github.com/ResearchHub/researchhub-ghostscript/tree/main

PR #1358, opened 2023-09-14
Arxiv api autopull

lightninglu10:
Autopulling from Arxiv directly everyday


src/paper/tasks.py
```diff
@@ -750,98 +753,214 @@ def pull_biorxiv_papers():
...
+                    elif author_name_length > 2:
+                        last_name = author_split[-1]
+                        first_name = ""
+                        for name_index, name_part in enumerate(author_split):
+                            if name_index != author_name_length - 1:
+                                first_name += name_part + " "
+                        first_name = first_name.strip()
+                    raw_authors.append(
+                        {"first_name": first_name, "last_name": last_name}
+                    )
```
le-sun, line 828:
```suggestion
        raw_authors = [
            {
                "first_name": " ".join(author_name[:-1]),
                "last_name": "".join(author_name[-1]),
            }
            for author in result.authors
            if (author_name := author.name.split(" "))
        ]
```

src/paper/tasks.py
```diff
@@ -750,98 +753,214 @@ def pull_biorxiv_papers():
...
+                "url": result.entry_id,
+                "raw_authors": format_raw_authors(raw_authors),
+                "title": title,
+                "paper_title": title,
+                "paper_publish_date": publication_date,
+                "is_open_access": True,
+                "oa_status": "",
+                "external_source": "arXiv",
+                "abstract": abstract,
+                "pdf_url": "https://arxiv.org/pdf/{}.pdf".format(entry_id),
```
le-sun, line 844:
```suggestion
                "pdf_url": f"https://arxiv.org/pdf/{entry_id}.pdf",
```

src/paper/tasks.py
```diff
@@ -750,98 +753,214 @@ def pull_biorxiv_papers():
...
-        arxiv_works = open_alex.get_data_from_source(arxiv_id, yesterday, page=i + 1)
+
+            data = {
+                "doi": pure_doi,
+                "url": result.entry_id,
+                "raw_authors": format_raw_authors(raw_authors),
+                "title": title,
+                "paper_title": title,
+                "paper_publish_date": publication_date,
+                "is_open_access": True,
```
le-sun, line 840: Hard coding open access feels iffy without a license check.

le-sun (changes requested):

src/paper/tasks.py
lightninglu10: There's no license info unfortunately in the API (or at least I couldn't find it). 

I think is_open_access is always true here -- it's just some articles may have a different open_access status. But their api for some reason doesn't expose this.

PR #1362, opened 2023-09-26
Get biorxiv tweet count

lightninglu10:
Ping the API to get biorxiv tweet count


src/paper/tasks.py
```diff
@@ -995,6 +995,24 @@ def pull_biorxiv(page=0, retry=0):
...
+        )
+        unescaped_json_res = res.text
+        # For some reason, the request data comes back with "\ufeff" and then also has an open paren ( and a closed paren )
+        escaped_json_res = json.loads(res.text.strip("\ufeff(").rstrip(res.text[-1]))
+        count = 0
+        for message in escaped_json_res["messages"]:
+            count += message["count_tweets"] + message["count_retweets"]
+        return count
+    except Exception as e:
+        print(e)
```
le-sun, line 1011: Delete

src/paper/tasks.py
```diff
@@ -995,6 +995,24 @@ def pull_biorxiv(page=0, retry=0):
         sentry.log_error(e)
 
 
+def _get_biorxiv_tweet_counts(url, doi):
+    try:
+        res = requests.get(
```
le-sun, line 1000: Let's add a timeout

le-sun (changes requested):
Couple small changes

src/paper/tasks.py
lightninglu10: I'd like to keep it in here, otherwise during development these errors are swallowed up

lightninglu10: better to print it to the console IMO

PR #1364, opened 2023-09-28
Tweet Scores refresh

lightninglu10:
Endpoint to refresh tweet scores

### DIFF
--- a/src/paper/tasks.py
+++ b/src/paper/tasks.py
@@ -927,11 +927,13 @@ def pull_biorxiv(page=0, retry=0):
 @app.task(queue=QUEUE_PULL_PAPERS)
 def set_biorxiv_tweet_count(url, doi, paper_id):
     from paper.models import Paper
+    from researchhub_document.signals import sync_scores
 
     counts = _get_biorxiv_tweet_counts(url, doi)
     paper = Paper.objects.get(id=paper_id)
     paper.twitter_score = counts
     paper.save()
+    sync_scores(paper.unified_document, paper)
 
 
 def _get_biorxiv_tweet_counts(url, doi):
--- a/src/paper/tasks.py
+++ b/src/paper/tasks.py
@@ -880,7 +880,7 @@ def get_biorxiv_tweets():
     biorxiv_papers = Paper.objects.filter(
         external_source__icontains="bioRxiv", created_date__gte=three_days_ago
     )
-    for paper in biorxiv_papers:
+    for paper in biorxiv_papers.iterator():
         set_biorxiv_tweet_count.apply_async(
             (
                 paper.url,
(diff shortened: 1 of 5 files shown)


src/paper/tasks.py
```diff
@@ -870,97 +870,26 @@ def pull_arxiv_papers_directly():
...
+    run_every=crontab(minute=0, hour="2"), priority=3, queue=QUEUE_PULL_PAPERS
+)
+def get_biorxiv_tweets():
+    from paper.models import Paper
+
+    three_days_ago = datetime.now() - timedelta(days=3)
+    biorxiv_papers = Paper.objects.filter(
+        external_source__icontains="bioRxiv", created_date__gte=three_days_ago
+    )
+    for paper in biorxiv_papers:
```
le-sun, line 883:
```suggestion
    for paper in biorxiv_papers.iterator():
```

src/paper/tasks.py
```diff
@@ -870,97 +870,26 @@ def pull_arxiv_papers_directly():
...
-#                     sentry.log_error(e)
-#         arxiv_works = open_alex.get_data_from_source(arxiv_id, yesterday, page=i + 1)
-#     reset_unified_document_cache(
-#         hub_ids=hub_ids,
-#         document_type=["paper"],
-#         filters=[NEW],
-#     )
-#     return total_works
+@periodic_task(
+    run_every=crontab(minute=0, hour="2"), priority=3, queue=QUEUE_PULL_PAPERS
```
le-sun, line 874: Maybe this should be daily instead of every 2 hours? It puts a lot of strain on our worker server

src/paper/views/paper_views.py
```diff
@@ -402,6 +403,17 @@ def censor(self, request, pk=None):
 
         return Response("Paper was deleted.", status=200)
 
+    @action(
+        detail=True,
+        methods=["post"],
+        permission_classes=[],
+    )
+    def update_biorxiv_tweet_count(self, request, pk=None):
+        paper = self.get_object()
```
le-sun, line 412: Can we add a check here to ignore updating the tweet count if it has been updated recently?

le-sun (changes requested):

src/paper/tasks.py
lightninglu10: This is Cronjob for every day at 2am right?

src/paper/views/paper_views.py
lightninglu10: Yes, we will also no longer be calling this endpoint

src/paper/tasks.py
le-sun: Oh yeah, I misread it.

PR #1368, opened 2023-10-04
Replication Prediction Market (No RSC)

manveerxyz:
Implement Replication Prediction Markets, without RSC voting (RSC voting is implemented in [manveer/v0-repl-market-rsc](https://github.com/manveerxyz/researchhub-backend/tree/manveer/v0-repl-market-rsc))

### Technical Details
- 2 new models:
  - `PredictionMarket`: Generic object meant for any prediction market.
    - `prediction_type`: currently only supports `REPLICATION_PREDICTION`
    - `status`: `OPEN`, `CLOSED`, or `RESOLVED`
  - `PredictionMarketVote`: A vote of `true` or `false` with optional `bet_amount`
- Return prediction market details in `get_document_metadata` endpoint
- When creating a vote on `POST` `/prediction_market_vote`:
  - Creates new `PredictionMarket` for paper if not yet exists
  - Updates user vote if user already voted
- Added tests for both prediction market and vote views


src/prediction_market/serializers/prediction_market_serializer.py
```diff
@@ -0,0 +1,47 @@
...
+        total_votes = obj.prediction_market_votes.count()
+        votes_for = obj.prediction_market_votes.filter(vote=True).count()
+        votes_against = total_votes - votes_for
+
+        return {"total": total_votes, "yes": votes_for, "no": votes_against}
+
+    def get_bets(self, obj):
+        # sum bets for and against
+        bets_for = (
+            obj.prediction_market_votes.filter(
```
lightninglu10, line 32: This will get slow really quickly. We should instead save this into the DB, probably under the PredictionMarket model and every time someone makes a new prediction vote, then update the stored value for that PredictionMarket.

src/prediction_market/serializers/prediction_market_serializer.py
```diff
@@ -0,0 +1,47 @@
...
+class DynamicPredictionMarketSerializer(DynamicModelFieldSerializer):
+    votes = SerializerMethodField()
+    bets = SerializerMethodField()
+
+    class Meta:
+        model = PredictionMarket
+        fields = ["id", "prediction_type", "end_date", "status", "votes", "bets"]
+
+    def get_votes(self, obj):
+        total_votes = obj.prediction_market_votes.count()
```
lightninglu10, line 23: We should also probably store this value in the PredictionMarket class to effectively "cache" it for faster loads

src/prediction_market/views/pred_market_view.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            prediction_market = create_prediction_market(paper_id)
+        except Exception as e:
+            return Response({"message": str(e)}, status=400)
+
+        data = DynamicPredictionMarketSerializer(prediction_market).data
+        return Response(data, status=200)
+
+    def retrieve(self, request, *args, **kwargs):
```
lightninglu10, line 42: In the ModelViewSet, the retrieve functionality is already predefined. Do we need to have an override here? Is it to define the DynamicPredictionMarketSerializer in order to remove specific fields from being returned?

src/prediction_market/views/pred_market_view.py
```diff
@@ -0,0 +1,45 @@
...
+from prediction_market.serializers.prediction_market_serializer import (
+    DynamicPredictionMarketSerializer,
+    PredictionMarketSerializer,
+)
+from prediction_market.utils import create_prediction_market
+from utils.throttles import THROTTLE_CLASSES
+
+
+class PredictionMarketViewSet(viewsets.ModelViewSet):
+    serializer_class = PredictionMarketSerializer
```
lightninglu10, line 16: If we store the values of votes / bets of the PredictionMarket inside the model itself, we'll need a way for users to not be able to edit that model.

src/prediction_market/views/pred_market_votes_view.py
```diff
@@ -0,0 +1,181 @@
...
+                    "created_date",
+                    "updated_date",
+                    "profile_image",
+                )
+            },
+        }
+        return context
+
+    def list(self, request, *args, **kwargs):
+        pred_mkt_id = request.query_params.get("prediction_market_id")
```
lightninglu10, line 125: In a ModelViewSet the list functionality is built in. 

If we just want to query by custom query parameters, you can look at django-filters (we use it through the rest of our projects too). 

Makes the code a bit easier to understand by fitting into that format

src/prediction_market/views/pred_market_votes_view.py
```diff
@@ -0,0 +1,181 @@
...
+            queryset = queryset.order_by(*ordering)
+
+        context = self._get_retrieve_context()
+        serializer = DynamicPredictionMarketVoteSerializer(
+            queryset, many=True, context=context
+        )
+        return Response(serializer.data)
+
+    def destroy(self, request, *args, **kwargs):
+        user = request.user
```
lightninglu10, line 155: Instead of this being a true delete, would it be better to have a status="neutral" or something on the vote? That way it's more of a "soft delete" and also keeps one item in the DB for a "Yes", "No", or "Remove Vote"

src/prediction_market/related_models/prediction_market_vote_model.py
```diff
@@ -0,0 +1,33 @@
...
+
+    prediction_market = models.ForeignKey(
+        "prediction_market.PredictionMarket",
+        related_name="prediction_market_votes",
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+    )
+
+    vote = models.BooleanField(
```
lightninglu10, line 23: Linking back to a comment below, maybe this should be an enum instead of a True / False boolean so we can capture the state of 

"yes", "no", "neutral" (removed vote)

src/prediction_market/views/pred_market_view.py
manveerxyz: Good catch, removed.

src/prediction_market/views/pred_market_view.py
manveerxyz: Set `permission_classes` to only allow Moderators, and overwrote any write/update methods to return `405: Not Implemented`

src/prediction_market/views/pred_market_votes_view.py
manveerxyz: Good point! Made it a soft delete.

src/prediction_market/serializers/prediction_market_serializer.py
manveerxyz: I moved them to DB fields, and setup signals to update the values when new votes are created/updated/soft-deleted — that seemed like the best design pattern.

src/prediction_market/views/pred_market_view.py
lightninglu10: If the permission_classes there is only moderators, will regular people be able to vote Yes or No?

manveerxyz: Yes, there's a separate view for votes with `IsAuthenticatedOrReadOnly`

PR #1369, opened 2023-10-05
Various fixes for hubs

yattias:
- Call openalex in URL flow
- Do not fail hub creation if a single concept in a list has problem


src/paper/paper_upload_tasks.py
```diff
@@ -650,12 +655,28 @@ def celery_create_paper(self, celery_data):
...
+                e,
+                message={
+                    "concept": concept,
+                    "concept.hub": concept.hub,
+                    "paper_concept": paper_concept,
+                },
+            )
+
+    if len(hubs_to_tag_paper_with) > 0:
+        paper.unified_document.hubs.add(*hubs_to_tag_paper_with)
```
le-sun, line 679: Can we add the hubs to the legacy paper hub too?

src/user/consumers.py
```diff
@@ -114,8 +114,14 @@ async def notify_paper_submission_status(self, event):
         # Not entirely sure if the paper submission serializer requires async support
         serialized_data = PaperSubmissionSerializer(submission).data
         paper = await get_paper_from_submission(submission)
-        unified_document = await get_unified_document_from_paper(paper)
-        hubs = await get_hubs_from_unified_document(unified_document)
+
+        hubs = []
+        unified_document = None
+
+        if paper:
```
le-sun, line 121: Was there some issue with the unified document being null?

le-sun (changes requested):

src/user/consumers.py
yattias: The issue is that paper is actually `None` at some point in this process. As a result, paper.unified_document fails

src/paper/paper_upload_tasks.py
yattias: yeah

PR #1371, opened 2023-10-09
Automated biorxiv bounties

lightninglu10:
Automate biorxiv bounties if they get above 10 upvotes

### DIFF
--- a/src/discussion/reaction_views.py
+++ b/src/discussion/reaction_views.py
@@ -5,7 +5,6 @@ from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
-from sentry_sdk import capture_exception
 
 from analytics.amplitude import track_event
 from discussion.permissions import CensorDiscussion as CensorDiscussionPermission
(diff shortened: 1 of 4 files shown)


src/discussion/reaction_views.py
```diff
@@ -406,6 +473,13 @@ def update_or_create_vote(request, user, item, vote_type):
 
     item.save()
 
+    try:
+        # If we're in the biorxiv review hub, we want all papers with 10 upvotes to get an automatic peer review
+        create_automated_bounty(item)
+    except Exception as e:
+        print(e)
+        capture_exception(e)
```
le-sun, line 481: Let's use `from utils.sentry import log_error` here

src/researchhub/settings.py
```diff
@@ -527,9 +527,9 @@ def silky_capture(request):
     "AWS_STORAGE_BUCKET_NAME", "researchhub-paper-dev1"
 )
 
-if PRODUCTION:
-    AWS_STORAGE_BUCKET_NAME = "researchhub-paper-prod"
-    AWS_S3_CUSTOM_DOMAIN = "d2ogkcqdn9wsvr.cloudfront.net"
+# if PRODUCTION:
+AWS_STORAGE_BUCKET_NAME = "researchhub-paper-prod"
+AWS_S3_CUSTOM_DOMAIN = "d2ogkcqdn9wsvr.cloudfront.net"
```
le-sun, line 532: Won't this break staging/dev?

le-sun (changes requested):

src/researchhub/settings.py
lightninglu10: yep, need to update

PR #1372, opened 2023-10-10
Allow users to request deletes and also delete their own accounts

lightninglu10:


yattias:
@lightninglu10 can we add a confirmation when user deletes their account? Wasn't sure if we have one in

lightninglu10:
Yeah, there is a confirmation step on the frontend

PR #1375, opened 2023-10-17
Bounty type

lightninglu10:
Adding bounty type to comments


yattias:
@lightninglu10 While at it, do you think you can fix the following issue:
If a peer review is removed, the peer review rating still remains at the DocumentHeader. I think the endpoint:
`/api/researchhub_unified_document/{id}/get_document_metadata/` needs to have the cache cleared.

lightninglu10:
> @lightninglu10 While at it, do you think you can fix the following issue: If a peer review is removed, the peer review rating still remains at the DocumentHeader. I think the endpoint: `/api/researchhub_unified_document/{id}/get_document_metadata/` needs to have the cache cleared.

Will do this in a separate PR

PR #1379, opened 2023-10-25
Verification

le-sun:
Account verification with Orcid and Linkedin
Code cleanup


src/hub/filters.py
```diff
@@ -1,55 +1,12 @@
...
-from django.utils import timezone
+from django.db.models import FileField
 from django_filters import rest_framework as filters
 
-from discussion.reaction_models import Vote
-
 from .models import Hub
 
 
-class ScoreOrderingFilter(filters.OrderingFilter):
```
lightninglu10, line 13: Did this filter get moved, or did we remove it entirely?

le-sun: Was testing something and forgot to put it back in.

PR #1380, opened 2023-10-26
Update Pagination Logic on Hubs

manveerxyz:
Had to make some changes to the backend to make pagination work properly.

- Ensure the `ordering` param is actually getting used
- Add "id" as an ordering field to break ties for hubs that have identical scores
- Added tests for ordering and pagination

### DIFF
--- a/src/hub/filters.py
+++ b/src/hub/filters.py
@@ -1,55 +1,11 @@
-from datetime import timedelta
-
 from django.contrib.postgres.search import TrigramSimilarity
-from django.db.models import Count, F, FileField, Q
-from django.utils import timezone
+from django.db.models import FileField
 from django_filters import rest_framework as filters
-
-from discussion.reaction_models import Vote
 
 from .models import Hub
 
-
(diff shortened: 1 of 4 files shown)


src/hub/filters.py
```diff
@@ -40,7 +40,7 @@ def filter(self, qs, value):
                 + F("paper_count")
             )
 
-            qs = qs.annotate(score=score).order_by("-score")
+            qs = qs.annotate(score=score).order_by(*value, "id")
```
lightninglu10, line 43: so we're ordering first by score and then by id?

manveerxyz: `id` is meant to be a tie-breaker if the scores are the same. This change is to resolve the issue we had with the same item showing up on multiple pages.

manveerxyz:
Okay, closed https://github.com/ResearchHub/researchhub-backend/pull/1381 and removed sorting filters in this PR.

Also updated the caching in the view.

lightninglu10:
Looks good, going to merge & deploy

PR #1381, opened 2023-10-27
Exclude votes from hubs popularity score

manveerxyz:
Our hubs endpoint was taking a long time to compute scores because it fetched upvotes/downvotes for the past two weeks across all papers in the hub.

We opted to remove the votes from the final score count for simplicity.

Alternatives considered:
- Reducing two week threshold down
- Setting up a cron job that periodically updated votes (perhaps for top X hubs)
- Setup signals for votes that auto-updates hub score

### DIFF
--- a/src/hub/filters.py
+++ b/src/hub/filters.py
@@ -1,11 +1,7 @@
-from datetime import timedelta
-
 from django.contrib.postgres.search import TrigramSimilarity
-from django.db.models import Count, F, FileField, Q
-from django.utils import timezone
+from django.db.models import F, FileField
 from django_filters import rest_framework as filters
 
-from discussion.reaction_models import Vote
 
 from .models import Hub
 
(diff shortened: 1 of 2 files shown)


src/hub/filters.py
```diff
@@ -17,26 +13,9 @@ def __init__(self, *args, **kwargs):
 
     def filter(self, qs, value):
         if value and any(v in ["score", "-score"] for v in value):
-            two_weeks_ago = timezone.now().date() - timedelta(days=14)
```
lightninglu10, line 20: instead of doing the filter here, can we remove this filter entirely? and change the `ordering=-paper_count / discussion_count`

lightninglu10: I think it would be

`?ordering=-paper_count,-discussion_count,-id`

manveerxyz: Does that even work? I'm trying that using `curl` but it's not working. It doesn't seem like Django supports complex ordering like that unless we have a custom Ordering Filter.

PR #1384, opened 2023-11-03
Citation metadata improvements

lightninglu10:
Metadata improvements to citations

### DIFF
--- a/src/utils/openalex.py
+++ b/src/utils/openalex.py
@@ -51,6 +51,40 @@ class OpenAlex:
                 data["works"] = response.json().get("results", [])
         return data
 
+    def map_to_csl_format(self, input_json, mapping=None):
+        """
+        Maps fields in the input_json based on the given mapping.
+
+        Args:
+            input_json (dict): JSON whose fields need to be mapped.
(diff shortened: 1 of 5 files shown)


src/utils/openalex.py
```diff
@@ -51,6 +52,44 @@ def _get_works_from_api_url(self, data):
...
+            # Check if the current key is in the mapping
+            if key in mapping:
+                print(key)
+                mapped_key = mapping[key]
+
+                # If the mapped key is a dictionary, recursively apply mapping to the input JSON value
+                if isinstance(mapped_key, dict) and isinstance(value, dict):
+                    output_json.update(self.map_to_csl_format(value, mapped_key))
+                else:
+                    if key == "abstract_inverted_index" or key == "abstract":
```
le-sun, line 82: We should add a check if the key is abstract and the value is not a string

le-sun (changes requested):
One small request

src/utils/openalex.py
lightninglu10: done

PR #1387, opened 2023-11-08
DOI uploads fetch PDFs

lightninglu10:
Fetch PDFs on DOI upload

### DIFF
--- a/src/citation/constants.py
+++ b/src/citation/constants.py
@@ -156,6 +156,7 @@ CITATION_TYPE_FIELDS = {
         "issued",
         "note",
         "journal-name",
+        "pdf_url",
     ],
     ARTWORK: [
         "title",
--- a/src/citation/constants.py
+++ b/src/citation/constants.py
@@ -156,7 +156,7 @@ CITATION_TYPE_FIELDS = {
         "issued",
         "note",
         "journal-name",
-        "pdf_url",
+        "custom",
     ],
     ARTWORK: [
         "title",
--- a/src/citation/schema.py
+++ b/src/citation/schema.py
@@ -156,10 +156,11 @@ OPENALEX_TO_CSL_FORMAT = {
     "abstract": "abstract",
     "landing_page_url": "url",
     "primary_location": {
+        "pdf_url": "pdf_url",
         "source": {
             "display_name": "source",
             "issn_l": "issn_l",
-        }
+        },
     },
 }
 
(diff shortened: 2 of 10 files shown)


src/citation/constants.py
```diff
@@ -156,6 +156,7 @@
         "issued",
         "note",
         "journal-name",
+        "pdf_url",
```
le-sun, line 159: Lets change this to `"URL"` because `"pdf_url"` is not CSL compliant.

src/citation/schema.py
```diff
@@ -156,10 +156,11 @@ def generate_schema_for_citation(citation_type):
     "abstract": "abstract",
     "landing_page_url": "url",
     "primary_location": {
+        "pdf_url": "pdf_url",
```
le-sun, line 159: Related to above comment

src/utils/openalex.py
```diff
@@ -1,9 +1,42 @@
...
+from paper.utils import get_pdf_from_url
 from researchhub.settings import OPENALEX_KEY
+from utils.aws import lambda_compress_and_linearize_pdf, upload_to_s3
+from utils.http import check_url_contains_pdf
 from utils.parsers import rebuild_sentence_from_inverted_index
 from utils.retryable_requests import retryable_requests_session
+from utils.sentry import log_error
+
+
+def download_pdf(url):
```
le-sun, line 14: Let's move this function somewhere more appropriate

le-sun (changes requested):

src/citation/constants.py
lightninglu10: we already have another field "URL" -- what should we do about that other field?

lightninglu10: The CSL fields are slightly limiting, I'm thinking the best way we can handle it is if we can transform our own "fields" JSON into CSL fields with a function, that way we can discard fields when we need a citation but can keep other metadata.

le-sun: CSL has the additional `"custom"` field, where we can enter our own key-value pairs

lightninglu10: ah, we should add it in there then. Is "custom" a JSON field?

le-sun: Yep.
In `schema.py`
```
            "custom": {
                "title": "Custom key-value pairs.",
                "type": "object",
                "description": "Used to store additional information that does not have a designated CSL JSON field. The custom field is preferred over the note field for storing custom data, particularly for storing key-value pairs, as the note field is used for user annotations in annotated bibliography styles.",
                "examples": [
                    {"short_id": "xyz", "other-ids": ["alternative-id"]},
                    {"metadata-double-checked": True},
                ],
            },
```

lightninglu10: Okay, made the change here. Will review again and deploy tomorrow

PR #1393, opened 2023-11-20
sweeper for depositing RSC

lightninglu10:
Checks every 5 minutes if we have open deposits and completes them if they are completed

### DIFF
--- a/src/reputation/tasks.py
+++ b/src/reputation/tasks.py
@@ -106,7 +106,7 @@ def check_transaction_success(transaction_hash):
 )
 def check_deposits():
     deposits = Deposit.objects.filter(paid_status=None)
-    for deposit in deposits:
+    for deposit in deposits.iterator():
         amt = deposit.amount
         user = deposit.user
         transaction_successful = check_transaction_success(deposit.transaction_hash)
--- a/src/reputation/tasks.py
+++ b/src/reputation/tasks.py
@@ -114,6 +114,7 @@ def check_deposits():
             distribution = Dist("DEPOSIT", amt, give_rep=False)
             distributor = Distributor(distribution, user, user, time.time(), user)
             distributor.distribute()
+            deposit.set_paid()
 
 
 @periodic_task(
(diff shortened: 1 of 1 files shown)


src/reputation/tasks.py
```diff
@@ -82,6 +82,40 @@ def create_author_contribution(contribution_type, user_id, unified_doc_id, objec
...
+
+
+@periodic_task(
+    run_every=crontab(minute="*/5"),
+    priority=4,
+    queue=QUEUE_PURCHASES,
+)
+def check_deposits():
+    deposits = Deposit.objects.filter(paid_status=None)
+    for deposit in deposits:
```
le-sun, line 109:
```suggestion
    for deposit in deposits.iterator():
```

src/reputation/tasks.py
```diff
@@ -82,6 +82,40 @@ def create_author_contribution(contribution_type, user_id, unified_doc_id, objec
...
+        amt = deposit.amount
+        user = deposit.user
+        transaction_successful = check_transaction_success(deposit.transaction_hash)
+        if transaction_successful:
+            distribution = Dist("DEPOSIT", amt, give_rep=False)
+            distributor = Distributor(distribution, user, user, time.time(), user)
+            distributor.distribute()
```
le-sun, line 116: Should we add a deposit update here to mark it as paid?
```suggestion
            distributor.distribute()
            deposit.set_paid()
```

le-sun (changes requested):

src/reputation/tasks.py
lightninglu10: Oh is the distribute not enough?

le-sun: The distributor just creates the distribution object and pays it out, but it doesn't update the deposit object.

lightninglu10: Ah you're right, done

le-sun (approved):

PR #1396, opened 2023-11-28
Add spammer status to author object

yattias:
Related to https://github.com/ResearchHub/researchhub-web/pull/1625

### DIFF
--- a/src/user/serializers.py
+++ b/src/user/serializers.py
@@ -61,7 +61,7 @@ class AuthorSerializer(ModelSerializer):
     total_score = SerializerMethodField()
     university = UniversitySerializer(required=False)
     wallet = SerializerMethodField()
-    status = SerializerMethodField()
+    suspended_status = SerializerMethodField()
 
     class Meta:
         model = Author
@@ -74,7 +74,7 @@ class AuthorSerializer(ModelSerializer):
             "num_posts",
             "orcid_id",
             "reputation",
-            "status",
+            "suspended_status",
             "sift_link",
             "total_score",
             "university",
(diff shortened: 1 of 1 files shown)


src/user/serializers.py
```diff
@@ -61,6 +61,7 @@ class AuthorSerializer(ModelSerializer):
     total_score = SerializerMethodField()
     university = UniversitySerializer(required=False)
     wallet = SerializerMethodField()
+    status = SerializerMethodField()
```
manveerxyz, line 64: just curious, why do we put this `status` on the `Author` and not the `User`?

manveerxyz (approved):

src/user/serializers.py
lightninglu10: I don't think we're returning the User object in the Author field

lightninglu10, line 64: IMO we should call this something else, like `suspended_status`

src/user/serializers.py
```diff
@@ -148,6 +150,12 @@ def get_num_posts(self, author):
             return ResearchhubPost.objects.filter(created_by=user).count()
         return 0
 
+    def get_status(self, author):
+        return {
```
lightninglu10, line 154: if this is `suspended_status`, then we should return a True / False here. So then the logic is dictated by the backend instead of being split between the FE / BE

src/user/serializers.py
yattias: yeah that sounds better

src/user/serializers.py
yattias: The main issue is that I am currently differentiating suspension from banned. Maybe we don't need to?

PR #1399, opened 2023-12-01
Fix error fetching prediction market votes

manveerxyz:
Jeff was unable to submit/retrieve prediction market votes.

I compared the retrieve context with the one in CommentView and realized it doesn't have the right fields/values set.

I also added a `try/catch` when getting `editor_of` details to make it more robust, and also because we have a similar `try/catch` when getting the author profile.

### DIFF
--- a/src/prediction_market/serializers/prediction_market_vote_serializer.py
+++ b/src/prediction_market/serializers/prediction_market_vote_serializer.py
@@ -28,7 +28,7 @@ class DynamicPredictionMarketVoteSerializer(DynamicModelFieldSerializer):
 
     def get_created_by(self, obj):
         context = self.context
-        _context_fields = context.get("rhc_dcs_get_created_by", {})
+        _context_fields = context.get("pm_dpmvs_get_created_by", {})
         serializer = DynamicUserSerializer(
             obj.created_by, context=context, **_context_fields
         )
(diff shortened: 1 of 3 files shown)


le-sun (changes requested):
Let's undo the `user/serializer.py` changes

The issue lies in the market vote dynamic serializer. 
In line 31 of `prediction_market_vote_serializer`,
```
_context_fields = context.get("rhc_dcs_get_created_by", {})
``` 
needs to be changed to 
```
_context_fields = context.get("pm_dpmvs_get_created_by", {})
```
This is because `rhc` stands for `researchhub comment` and `dynamic comment serializer` for `dcs`

Then in `pred_market_votes_view.py`,
line 160 nees to be changed to `"pm_dpmvs_get_created_by"` from `rhc_dcs_get_created_by`. You also need to get rid of line 166 `"editor_of",`

manveerxyz:
> Let's undo the `user/serializer.py` changes
> 
> The issue lies in the market vote dynamic serializer. In line 31 of `prediction_market_vote_serializer`,
> 
> ```
> _context_fields = context.get("rhc_dcs_get_created_by", {})
> ```
> 
> needs to be changed to
> 
> ```
> _context_fields = context.get("pm_dpmvs_get_created_by", {})
> ```
> 
> This is because `rhc` stands for `researchhub comment` and `dynamic comment serializer` for `dcs`
> 
> Then in `pred_market_votes_view.py`, line 160 nees to be changed to `"pm_dpmvs_get_created_by"` from `rhc_dcs_get_created_by`. You also need to get rid of line 166 `"editor_of",`

That makes so much more sense, I had no clue what the rhc/dcs stuff stood for. Made the changes!

PR #1400, opened 2023-12-05
Ref Manager Fixes

lightninglu10:
Resolves issues listed here:

Screenshot 2023-12-05 at 11 27 47 AM


src/citation/related_models/citation_entry_model.py
```diff
@@ -70,14 +70,18 @@ def organization_indexing(self):
...
-            org_permissions = self.organization.permissions
-            return org_permissions.has_editor_user(
-                user
-            ) or org_permissions.has_admin_user(user)
-        else:
-            project_permissions = belonging_project.permissions
-            return project_permissions.has_editor_user(
-                user
-            ) or project_permissions.has_admin_user(user)
+        return self.project.created_by == user or {"full_access": True}.get(
```
le-sun, line 73: I don't understand the permission changes

le-sun (changes requested):

src/citation/related_models/citation_entry_model.py
lightninglu10: I changed the permissions on the frontend. This is to match that. All users in the org have access if the project is full access, unless you're the project creator, then you always have access.

The old permissions were good but were broken and this is easier to fix for now.

PR #1403, opened 2023-12-08
Updates

yattias:


src/reputation/views/withdrawal_view.py
```diff
@@ -109,10 +110,10 @@ def oz_webhook(self, request):
...
 
     def create(self, request):
-        if timezone.now() < timezone.make_aware(timezone.datetime(2020, 9, 1)):
-            return Response(
-                "Withdrawals are disabled until September 1, 2020", status=400
-            )
+        if LogEntry.objects.filter(
+            object_repr="WITHDRAWAL_SWITCH", action_flag=3
+        ).exists():
+            return Response(status=503)
```
lightninglu10, line 116: In this response, can you return a message?

Like "Withdrawals are suspended for the time being. Please be patient as we work to turn withdrawals back on"

Something like this

lightninglu10 (approved):

PR #1404, opened 2023-12-13
Deduct fees when tipping RSC

manveerxyz:
Deduct fees when handling tipping purchases.

Leverages existing logic from bounty fees.


yattias (commented):
@manveerxyz code looks good. Before we merge it though, I think we should write a couple of tests:
1. Test that validates that a commission transaction is created
"Should create commission transaction"

2. Test validates that tipping from one user to another works
"Should tip another user"

PR #1405, opened 2023-12-15
Post in ref manager

lightninglu10:
Allows users to create posts in reference manager with a new endpoint


src/citation/schema.py
```diff
@@ -8,6 +8,7 @@
     ARTICLE_JOURNAL,
     BIBTEX_TO_CITATION_TYPES,
     BIBTEX_TYPE_TO_CSL_MAPPING,
+    BLOG_POST,
```
manveerxyz, line 11: Not 100% sure if this is the best type to use for posts? Based off of [CSL types](https://docs.citationstyles.org/en/stable/specification.html#appendix-iii-types), generic `article` seems to make sense.

If not, I think it'd still be good to have [the fields include DOI](https://github.com/ResearchHub/researchhub-backend/blob/fa05231ed7448d475bc0064291548b2e14cad826/src/citation/constants.py#L250) so that posts with DOIs have their DOI show in the metadata section on the reference manager.

src/citation/schema.py
```diff
@@ -56,6 +57,61 @@ def generate_json_for_doi_via_oa(doi):
...
+            elif field == "abstract":
+                value = post.renderable_text[0:255]
+                if len(post.renderable_text) > 255:
+                    value += "..."
+            elif field == "genre":
+                hub_names = post.hubs.values_list("name", flat=True)
+                hub_names_string = ", ".join(hub_names)
+                value = hub_names_string
+            elif field == "URL":
+                value = "www.researchhub.com/"
```
manveerxyz, line 108: Shouldn't this be a direct link to the post?

manveerxyz (commented):
Looks good for the most part! Just a few comments.

PR #1407, opened 2024-01-02
Verification via author claim

yattias:
- One-off script to update users who have previously claimed a paper as verified ~130 users
- Updated existing paper claim flow to support for claiming a paper without the paper being on RH. It is possible to claim a paper just via DOI
- New endpoint for searching for papers in OpenAlex via DOI
- Papers verified via DOI that are not on RH will be uploaded to RH
- Send In-app notification to users who have claimed papers and verified account


src/researchhub_case/related_models/author_claim_case_model.py
```diff
@@ -34,6 +34,8 @@ class AuthorClaimCase(AbstractResearchhubCase):
         on_delete=models.CASCADE,
         related_name="related_claim_cases",
     )
+    target_paper_doi = models.CharField(max_length=255, null=True)
+    target_paper_title = models.CharField(max_length=1024, null=True)
```
manveerxyz, line 38: nice attention to detail on `max_length`

src/user/management/commands/backfill_verified_users.py
```diff
@@ -0,0 +1,82 @@
...
+                # In-app notification about verification approval
+                verification_notification = Notification.objects.create(
+                    item=first_user_claim,
+                    notification_type=Notification.ACCOUNT_VERIFIED,
+                    recipient=user,
+                    action_user=user,
+                )
+
+                verification_notification.send_notification()
+                send_verification_email(first_user_claim, context={})
```
manveerxyz, line 80: this is not a big issue and I think it's fine as-is, but wanted to flag in-case you thought it'd merit change: it might be a little weird to get an email that I got verified with no explanation when I didn't do anything to trigger it.

manveerxyz (commented):
Backend looks pretty good to me.

src/user/management/commands/backfill_verified_users.py
yattias: That's a good point, I will add a reason to the email. I think it is important to get an email so the user isn't surprised by being verified but agree that they should know why the email has been triggered.

PR #1408, opened 2024-01-03
Setup Preregistration Crowdfunding

manveerxyz:
It's useful to go through this PR commit-by-commit.

## Move purchase views/serializers into folders
Self-explanatory, mainly for organizability.

## Setup new document type `PREREGISTRATION`
Self-explanatory.

## Implement Preregistration Fundraising
- New `Fundraise` model
  - Defaults `end_date` to 30 days from now
  - Uses `Escrow` to store RSC and handle payout
  - Uses `Purchase` to track contributions
- New `FundraiseViewSet`
  - Tried to ensure all edge-cases are accounted for (e.g. closed/expired fundraise, insufficient funds, etc.)
  - Triggers payout automatically once fundraise is fulfilled
  - Only allow moderators to create new fundraise using:
```bash
curl -X POST 'https://<BASE_URL>/api/fundraise/' -H 'Authorization: Token <TOKEN>' -H 'Content-Type: application/json' -d '{"goal_amount": <GOAL>, "goal_currency": "USD", "unified_document_id": <ID>, "recipient_user_id": <ID>}' | json_pp
```

## Add fundraise tests
Self explanatory, tried to handle main edge cases.

PR #1409, opened 2024-01-08
Reduce tipping fee to 3%

manveerxyz:
Created a new `SupportFee` object.

Migration should automatically add a row for the new fee.


src/purchase/views.py
manveerxyz, line 1: This file should've been removed in https://github.com/ResearchHub/researchhub-backend/pull/1408 but it wasn't. Might've been a rebase issue.

src/researchhub_document/migrations/0057_alter_researchhubpost_document_type_and_more.py
yattias, line 1: I think you have some fundraise stuff in here - which is fine. Just wanted to flag it to ensure we need this here

src/purchase/views/purchase_view.py
```diff
@@ -89,14 +89,14 @@ def create(self, request):
...
-                total_fee, rh_fee, dao_fee, current_bounty_fee = calculate_fees(
+                total_fee, rh_fee, dao_fee, current_support_fee = calculate_support_fees(
                     decimal_amount
                 )
                 if user_balance - (decimal_amount + total_fee) < 0:
                     return Response("Insufficient Funds", status=402)
                 
                 # Deduct fees from the gross amount of the purchase.
-                deduct_fees(user, total_fee, rh_fee, dao_fee, current_bounty_fee)
+                deduct_support_fees(user, total_fee, rh_fee, dao_fee, current_support_fee)
```
yattias, line 99: Nice job separating out the fees from bounties and tips

yattias (approved):

src/researchhub_document/migrations/0057_alter_researchhubpost_document_type_and_more.py
manveerxyz: Good catch this shouldn't be here.

PR #1410, opened 2024-01-10
Improve how we handle copyrighted PDFs

manveerxyz:
### Main Technical Decisions
- Hide `file`, `pdf_url`, `first_preview` in Paper serializers if PDFs are copyrighted
    - Made change in serializer so that any EP that fetches papers correctly hides PDFs
- Added new field `is_pdf_removed_by_moderator` for when moderators remove PDF


src/paper/tests/work_by_doi.json
manveerxyz, line 1: Updated this object with the latest version from OpenAlex API.

src/utils/tests/work_by_doi.json
manveerxyz, line 1: Updated this object with the latest version from OpenAlex API.

src/paper/tests/test_paper_submissions.py
manveerxyz, line 1: For some reason, this object's fields got updated in OpenAlex: [Link to updated object](https://api.openalex.org/works?filter=doi:10.34133/2020/8086309)

yattias (approved):
Code looks great

yattias:
@manveerxyz Code looks good. I ran into one bug:
1. Uploaded paper publicly https://www.nature.com/articles/s41586-023-06902-y
2. Set license => "Publisher specific" then noticed PDF is removed: Great ✅ 
3. Saved to my ref manager, then noticed PDF appears in there.

I think if license is set to a closed one, then PDF shouldn't appear in ref manager. Could be that I'm missing something?
Would be great to write a test for this.

Finally, noticed that tests in this branch are failing. I re-started the jobs just to confirm but seems like might be legitimate failure

manveerxyz:
> @manveerxyz Code looks good. I ran into one bug:
> 
> 1. Uploaded paper publicly https://www.nature.com/articles/s41586-023-06902-y
> 2. Set license => "Publisher specific" then noticed PDF is removed: Great ✅
> 3. Saved to my ref manager, then noticed PDF appears in there.
> 
> I think if license is set to a closed one, then PDF shouldn't appear in ref manager. Could be that I'm missing something? Would be great to write a test for this.
> 
> Finally, noticed that tests in this branch are failing. I re-started the jobs just to confirm but seems like might be legitimate failure

1. Fixed the tests
2. I think my latest branch wasn't deployed to staging as I tested locally and it worked okay. I deployed the latest version so it should work now.

PR #1413, opened 2024-01-17
Implement Amplitude Revenue Tracking

manveerxyz:
Created a separate `track_revenue_event` that we should call whenever we earn revenue.

Will have to wait til it's in prod to test whether it fires correctly. 

Tested:
- Manually logged data to make sure we're tracking accurate/correct numbers
- Tipping still works
- Creating bounties still works

### DIFF
--- a/src/analytics/amplitude.py
+++ b/src/analytics/amplitude.py
@@ -2,10 +2,8 @@ import functools
 import json
 
 import requests
-from typing import Literal
 from ipware import get_client_ip
 
-from purchase.related_models.rsc_exchange_rate_model import RscExchangeRate
 from researchhub.settings import AMPLITUDE_API_KEY, DEVELOPMENT
 from utils.parsers import json_serial
 from utils.sentry import log_info
@@ -102,7 +100,7 @@ class Amplitude:
         hit = json.dumps(hit, default=json_serial)
         return self.forward_event(hit)
     
-    def track_revenue_event(
+    def _track_revenue_event(
         self,
         user,
         revenue_type: str,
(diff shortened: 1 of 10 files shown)


src/analytics/amplitude.py
```diff
@@ -99,6 +101,36 @@ def build_hit(self, res, view, request, *args, **kwargs):
         }
         hit = json.dumps(hit, default=json_serial)
         return self.forward_event(hit)
+    
+    def track_revenue_event(
+        self,
+        user,
+        revenue_type: str,
```
yattias, line 108: Not sure if Django allows for constants here

src/analytics/amplitude.py
```diff
@@ -124,3 +156,56 @@ def inner(*args, **kwargs):
         return res
 
     return inner
+
+
+def track_revenue_event(
+    user,
+    revenue_type: Literal["BOUNTY_FEE", "FUNDRAISE_CONTRIBUTION_FEE", "SUPPORT_FEE", "DOI_FEE", "WITHDRAWAL_FEE"],
```
yattias, line 163: Nice use of literals in the function signature. Didn't know this was possible

src/analytics/amplitude.py
```diff
@@ -124,3 +156,56 @@ def inner(*args, **kwargs):
...
+        if usd_revenue is None:
+            if not isinstance(rsc_revenue, float):
+                rsc_revenue_float = float(rsc_revenue)
+            else:
+                rsc_revenue_float = rsc_revenue
+
+            usd_revenue_float = RscExchangeRate.rsc_to_usd(rsc_revenue_float)
+            usd_revenue = str(usd_revenue_float)
+    except Exception as e:
+        log_info("Error converting RSC to USD", e)
```
yattias, line 195: Wondering if we should log error to sentry here instead of info so we get notified in slack

src/analytics/amplitude.py
```diff
@@ -99,6 +101,36 @@ def build_hit(self, res, view, request, *args, **kwargs):
         }
         hit = json.dumps(hit, default=json_serial)
         return self.forward_event(hit)
+    
+    def track_revenue_event(
```
yattias, line 105: I would add a try/except block in the definition here because this is used in critical flows of the app and if this throws an exception so will the calling function.

PR #1414, opened 2024-01-19
Reduce Bounty Minimum and Update Support Fee Allocation

manveerxyz:
- Reduce bounty minimum to 10 RSC
- Update tipping fee to give 3% to RH Inc. and 0% to DAO


src/reputation/utils.py
```diff
@@ -93,6 +93,13 @@ def _deduct_fees(
         giver=user,
     )
     rh_inc_record = rh_inc_distributor.distribute()
+
+    # if there's no dao fee, we can skip that distribution
+    if dao_fee == 0:
```
yattias, line 98: is `rh_inc_record` related to `dao_fee` ? if so, for consistency, can we name `rh_dao_record` ?

manveerxyz: oh this is kind of confusing, there's some logic a couple of lines below that throws an error if `rh_inc_record` and/or `rh_dao_record` are undefined, so if I were to return early I'd need to check if `rh_inc_record` is defined or not.

I'm going to just refactor this to be more understandable

PR #1416, opened 2024-01-25
Scripts to backfill OpenAlex and clean-up data

manveerxyz:
# Cleanup Scripts

### Fix Bad DOIs
Some DOIs aren't properly formatted:
- Base64 string
- Missing host id prefix

So fix them intelligently using `url` field.

### Fix Empty Titles/Dates
There's some papers with empty titles and/or dates, but OpenAlex has that data so just supplement the db rows with OpenAlex data.

### Remove Duplicates by DOI
Some DOIs in db have `https://doi.org` or `doi.org` prefix, so normalize these and find duplicates, then remove duplicate if there's no discussion.

### Update Abstracts
Update papers with missing abstracts by parsing `abstract_inverted_index` in OpenAlex data.

### Supplement with OpenAlex
Supplement papers with non-zero discussion_count with OpenAlex data.

# Backfill Scripts

### Rough Steps
1. Download OpenAlex snapshot (>330GB)
2. Use script to parse/process data and create CSV files
3. Bulk-load CSV files into temp tables in database
4. Update database tables with new data

### More details
1. OpenAlex Snapshot
    - A bunch of compressed `.jsonl` files
    - Only organized by when they were last updated in OpenAlex
2. Processing Scripts
    - Script goes through _all_ compressed files and filters for papers published in last 5 years
    - Excludes objects that aren't articles (e.g. grants, datasets, etc.)
    - Transforms OpenAlex data to our db schema
    - Creates CSV files for `paper`, `unified_document`, and `hubs` / `concepts` data
    - CSV files are split into chunks by year for easier processing
    - Deletes original files after processing to preserve disk space
3. Bulk-load CSV
    - Use `psql` to load CSV files into temp tables for processing
        - Bulk-loading CSV files is the most efficient way to import large data
4. Update tables
    - Add/update entries in the `paper_paper` table
    - Create new `unified_document` entries for all new papers
    - Go back and update the new `paper_paper` entries with new `unified_document_id`
    - Insert document concepts and hubs

---
### Alternatives considered
- Setup a job that queries OpenAlex API
    - Would be much slower and less efficient
- Create python scripts that leverage Django data model and use that to insert data with high data integrity
    - Would also be much slower, and the data model isn't too complex for raw SQL commands

### Concerns
- Large database inserts can lock tables for long amounts of time
    - Will test in staging to see how long these inserts take, if they take a while we can chunk them up

### DIFF
--- a/scripts/backfill_openalex.py
+++ b/scripts/backfill_openalex.py
@@ -1,0 +1,353 @@
+import gzip
+import json
+import pandas as pd
+import os
+from datetime import datetime
+from pathlib import Path
+from unicodedata import normalize
+from slugify import slugify
+import secrets
+import shutil
+
+def check_has_enough_disk_space():
+    """
+    Check if the machine has enough disk space to continue.
+    """
+    # 5 GB
+    min_free_space = 5 * 1024 * 1024 * 1024
+    path_to_check = '/'
+
+    total, used, free = shutil.disk_usage(path_to_check)
+
+    total_in_gb = total / (1024 * 1024 * 1024)
+    used_in_gb = used / (1024 * 1024 * 1024)
+    free_in_gb = free / (1024 * 1024 * 1024)
+
+    print("Disk space:")
+    # Print with only 2 decimal places
+    print(f"\tTotal: {total_in_gb:.2f} GB")
+    print(f"\tUsed: {used_in_gb:.2f} GB")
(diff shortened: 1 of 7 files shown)


yattias:
@manveerxyz this looks really solid. I like the approach. My concern is similar to yours - what happens if we have a DB lock when doing the mega inserts. Let's see how it goes on staging.

yattias (approved):

PR #1417, opened 2024-01-25
WIP: Personalization

yattias:
- Exporting historical item data in format suitable for ML
- Exporting historical user interaction data in format suitable for ML
- Integrating w/Segment to log key events necessary for ML

Next up:
- Daily export of item data
- Log missing events and items

### Screenshot from Segment.io
image


src/user/related_models/action_model.py
```diff
@@ -54,16 +55,19 @@ def __str__(self):
...
-        if self.id is None:
-            from mailing_list.tasks import notify_immediate
+        from mailing_list.tasks import notify_immediate
+
+        is_new = self.id is None
+        super().save(*args, **kwargs)
+
+        if is_new:
+            print("newly created", self)
+            log_analytics_event(self)
```
yattias, line 65: Going to update this to a celery job before I merge

src/analytics/management/commands/export_interactions_to_personalize.py
```diff
@@ -0,0 +1,134 @@
...
+                "hubs",
+                "user__author_profile",
+            )
+        )
+
+        if start_date_str:
+            start_date = datetime.strptime(start_date_str, "%Y-%m-%d").date()
+            actions_queryset = actions_queryset.filter(created_date__gte=start_date)
+
+        print(f"Number of documents >= {start_date_str}: " + str(len(actions_queryset)))
```
yattias, line 119: Looks weird in github, but just an output statement to help provide some info

src/analytics/misc/amazon_personalize_user_interaction_schema.json
yattias, line 1: This isn't used in code. This is uploaded to Personalize. Committing this for reference

src/analytics/management/commands/export_items_to_personalize.py
```diff
@@ -0,0 +1,195 @@
...
+                record["created_by_user_id"] = str(specific_doc.created_by.id)
+
+            data.append(record)
+        except Exception as e:
+            print("Failed to export doc: " + str(doc.id), e)
+
+    return data
+
+
+def map_comment_data(comments):
```
manveerxyz, line 78: why don't we store the comment content here? is it because it gets too long?

src/analytics/utils/analytics_mapping_utils.py
```diff
@@ -0,0 +1,205 @@
...
+from datetime import datetime
+
+import segment.analytics as analytics
+
+from discussion.reaction_models import Vote
+from paper.utils import format_raw_authors
+from researchhub_comment.related_models.rh_comment_model import RhCommentModel
+
+
+def log_analytics_event(action):
```
manveerxyz, line 13: should tipping and contributing to a fundraise technically be tracked as user interactions too? perhaps it's something we have to implement in the future.

manveerxyz (commented):
I think it looks good for the most part, just some minor questions.

src/analytics/management/commands/export_items_to_personalize.py
yattias: According to documentation, it is not necessary. The important part are categorical information and user interaction data (how people engage with the comment)

src/analytics/utils/analytics_mapping_utils.py
yattias: Yeah, as a next step I am going to add more events such as tipping

PR #1419, opened 2024-01-31
Personalization export schema changes

yattias:
- Minor schema changes to Amazon Personalize export


src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -80,6 +82,29 @@ def get_queryset(self):
             return ResearchhubUnifiedDocument.all_objects.all()
         return super().get_queryset()
 
+    @action(
```
yattias, line 85: This is WIP and will be updated in the next day

PR #1423, opened 2024-02-02
Speed optimization for recs endpoint

yattias:
Updated recs endpoint to be more performant at fetching db records

### DIFF
--- a/src/researchhub_document/views/researchhub_unified_document_views.py
+++ b/src/researchhub_document/views/researchhub_unified_document_views.py
@@ -4,6 +4,7 @@ import boto3
 from dateutil import parser
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
+from django.db.models import Q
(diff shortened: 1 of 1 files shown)


src/researchhub_document/views/researchhub_unified_document_views.py
```diff
@@ -81,64 +82,48 @@ def get_queryset(self):
...
-        user_id = request.query_params.get("user_id", None)
+        user_id = request.query_params.get("user_id")
+        if not user_id:
+            return Response({"error": "user_id is required"}, status=400)
+
         personalize_runtime = boto3.client(
             "personalize-runtime", region_name="us-west-2"
         )
-
         campaign_arn = "arn:aws:personalize:us-west-2:794128250202:campaign/hp-recs"
```
yattias, line 94: This hard-coded value is temporary and safe to expose. This is a shim until we write a new API for mixed content in next sprint.

PR #1426, opened 2024-02-06
Batch process OpenAlex frontfill and leverage celery multi-tasking

manveerxyz:
Something happening with the OpenAlex job when we pushed it live, I suspect it crashed because of all the data being processed.

This PR attempts to batch process new OpenAlex works to ease the load, as well as add better edge-case handling.


src/paper/tasks.py
```diff
@@ -1308,107 +1307,38 @@ def pull_new_openalex_works(page=0, retry=0):
     except Exception as e:
         sentry.log_error(e)
 
+    total_papers_processed = 0
+    cursor = "*" # used for pagination
```
yattias, line 1311: Can we expand on this in a comment? I'm not sure how exactly this is used in pagination

src/paper/tasks.py
```diff
@@ -1308,107 +1307,38 @@ def pull_new_openalex_works(page=0, retry=0):
...
-        works = open_alex.get_new_works(since_date=fetch_since_date)
-        if not works or len(works) == 0:
-            PaperFetchLog.objects.create(
-                source=PaperFetchLog.OPENALEX,
-                fetch_type=PaperFetchLog.FETCH_NEW,
-                status=PaperFetchLog.SUCCESS,
-                started_date=start_date,
-                completed_date=datetime.now(),
+        
+        while True:
```
yattias, line 1315: Is there a chance of this not terminating?

src/paper/tasks.py
```diff
@@ -1308,107 +1307,38 @@ def pull_new_openalex_works(page=0, retry=0):
...
-                    ),
-                    priority=2,
+            for work in works:
+                _process_openalex_work.apply_async(
+                    (work,),
+                    priority=4,
                     countdown=1,
                 )
+                # delay so that we don't queue too many tasks at once
+                time.sleep(0.2)
```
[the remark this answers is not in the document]
manveerxyz: It's slightly different, countdown delays the execution of the task, whereas `time.sleep` delays the queuing.

I added it so that we wouldn't just be constantly fetching works and adding them to the queue right away, this `time.sleep` is meant to give the queue some time to run through the queued tasks.

I don't know how efficacious it'd even be tbh, but I wanted to add something to reduce load.

src/paper/tasks.py
manveerxyz: I think it's unlikely given I'm checking `next_cursor` and for null or empty works response.

src/paper/tasks.py
manveerxyz: Done!

yattias (approved):

PR #1427, opened 2024-02-08
Improve resiliency and batch OpenAlex frontfill

manveerxyz:
- Log more errors / messaging to Sentry
- Continue where left off if retrying the job
- Avoid initializing multiple pull jobs
- Batch queries to `paper_paper` table wherever possible


src/paper/tasks.py
```diff
@@ -1269,45 +1308,97 @@ def pull_openalex_author_works(user_id, openalex_id):
...
+    3. If we hit an error, retry the job from where we left off
+    4. Log the results
+    """
     if not PRODUCTION:
         return
 
     from paper.models import PaperFetchLog
 
-    sentry.log_info("Starting New OpenAlex pull")
+    fetch_since_date = datetime.now() - timedelta(days=1)
```
yattias, line 1331: What does `fetch_since_date` represent? I don't think I fully understand what it means. Is it meant to represent the last time we tried fetching aka `last_fetch_date` ?

yattias: I think I got it. It represent the last date we fetched papers from. Kinda minor, but can we change to `fetch_from_date` or `date_to_fetch_from` since due to the nature of SQL, it is a bit clearer what is happening.

src/paper/tasks.py
```diff
@@ -1269,45 +1308,97 @@ def pull_openalex_author_works(user_id, openalex_id):
...
+    # if paper_fetch_log_id is provided, it means we're retrying
+    # otherwise we're starting a new pull
+    if paper_fetch_log_id is None:
+        start_date = datetime.now()
 
-    start_date = datetime.now()
+        # figure out when we should start fetching from.
+        # if we have an existing successful run, we start from the last successful run
+        try:
+            last_log = PaperFetchLog.objects.filter(
```
yattias, line 1340: Can we change name to `last_successful_run_log` or something like that. Kinda verbose but it does help when reading this.

src/paper/tasks.py
```diff
@@ -1319,80 +1410,143 @@ def pull_new_openalex_works(page=0, retry=0):
...
+            # if we've reached the end of the results, exit the loop
             if next_cursor is None or works is None or len(works) == 0:
                 break
 
-            for work in works:
-                _process_openalex_work.apply_async(
-                    (work,),
-                    priority=4,
+            # if we're starting from a specific index, skip until we reach that index
+            if total_papers_processed >= start_index:
```
yattias, line 1422: This is fine. I guess there is no way to start at a particular index?

yattias (commented):
I think code looks good. Had few comments / questions

src/paper/tasks.py
manveerxyz: Updated!

src/paper/tasks.py
manveerxyz: Yup I like that, it's complicated code so good to have verbosity.

src/paper/tasks.py
manveerxyz: Updated it to account of the case where the start_index is inbetween a batch

src/paper/tasks.py
```diff
@@ -247,6 +248,44 @@ def add_orcid_authors(paper_id):
     return True
 
 
+@app.task(queue=QUEUE_PAPER_MISC)
+def add_orcid_authors_batch(paper_ids):
```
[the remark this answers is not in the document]
manveerxyz: that's a fair point, I kind of just copied the existing orcid authors add code and batched it.

I'm going to deploy this current version and then make the changes you suggested tomorrow, since this is technically existing logic that's been live for a while.

PR #1441, opened 2024-02-21
Create an endpoint for proxying requests to route around CORS issues

manveerxyz:
Create a `/proxy` route that we can use to avoid CORS issues when fetching PDFs.

### DIFF
--- a/src/proxy/views.py
+++ b/src/proxy/views.py
@@ -1,0 +1,51 @@
+# views.py
+
+import requests
+from django.http import HttpResponse, StreamingHttpResponse
+from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


manveerxyz:
Had a conversation with Kobe about moving this to a separate service, perhaps a lightweight EC2 instance.
Will investigate and close this PR if that's the path forward.

lightninglu10:
For what it's worth, agree with Kobe here. 

A route like this should also be protected with an api key or some auth as people could start sending requests through our server. There's a lot of concerns when opening up a proxy route like this that can be abused, esp since code is open source.

PR #1444, opened 2024-02-23
Reduce database load from Unified Doc fetch

manveerxyz:
## Context
There were a lot of `SELECT COUNT(*)` queries on Unified Document table, after investigating it became clear that basically all of them were from requests to the `get_unified_documents` endpoint.

When I logged the raw SQL run by Django, and tried requests to this endpoint locally, the raw SQL queries exactly matched the queries in the slow query log.

It seems like the `SELECT COUNT(*)` would get called from the underlying `Paginator` object when fetching the unified documents.

## Fix

We already had some logic to use a "cached" version of the row count for speed, but the underlying `Paginator` object was still calling the `SELECT COUNT(*)`.

I modified the custom paginator to correctly use the cached row count in all cases.

## Testing

Tested locally by logging the raw SQL that Django runs, and didn't see any `SELECT COUNT(*)` on the Unified Document table.

### DIFF
--- a/src/researchhub_document/views/custom/unified_document_pagination.py
+++ b/src/researchhub_document/views/custom/unified_document_pagination.py
@@ -1,31 +1,58 @@
 from collections import OrderedDict
 
 from django.db import connection
+from django.core.paginator import Paginator
+from django.utils.functional import cached_property
 from rest_framework.pagination import PageNumberPagination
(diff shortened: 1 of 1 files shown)


yattias (approved):
This looks promising, let's try it out

PR #1457, opened 2024-03-12
Fix document tests

gzurowski:
- Some of the test never ran because test methods were not prefixed with `test_`.
- Fix existing texts (update hypothesis, update post with different user than author).


yattias (approved):
LGTM

koutst (approved):
LGTM

PR #1464, opened 2024-03-15
Verify post authors

gzurowski:
Verify that all provided authors are all part of the organization under which the post is published. This check is added to the publish post and republish post use cases.

### DIFF
--- a/src/researchhub_document/views/researchhub_post_views.py
+++ b/src/researchhub_document/views/researchhub_post_views.py
@@ -207,8 +207,18 @@ class ResearchhubPostViewSet(ReactionViewActionMixin, ModelViewSet):
     @sift_track(SIFT_POST, is_update=True)
     def update_existing_researchhub_posts(self, request):
         data = request.data
+
+        authors = data.get("authors", [])
+        rh_post_id = data.get("post_id", None)
+        rh_post = ResearchhubPost.objects.get(id=rh_post_id)
+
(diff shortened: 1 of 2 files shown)


koutst (approved):
LGTM

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -103,41 +101,39 @@ def upsert_researchhub_posts(self, request):
         except (KeyError, TypeError) as exception:
             return Response(exception, status=400)
 
+    def _check_authors_in_org(self, authors, organization):
+        for author_id in authors:
+            author = User.objects.get(id=author_id)
+            if not organization.org_has_user(author):
+                return False
+        return True
```
yattias, line 109: I ran into an issue while testing. This line in particular
```
User.objects.get(id=author_id)
```

Throws an error because an `author_id` is passed instead of a user_id.
I think this line should be updated to  
```
Author.objects.get(id=author_id)
````

There is a 1:1 relationship between author and user. It seems like author ids are passed in the request and not user ids. This may have not errored out in your local env because locally, it is very likely that the author ids and user ids are identical since they are created sequentially

gzurowski: Thanks for pointing this out. Correct, I haven't noticed this in my local environment due to user IDs and author IDs aligning.

yattias (approved):
LGTM

PR #1467, opened 2024-03-18
Add RH document type migration

gzurowski:
The document type has been recently extended with #1408, but the corresponding database migration hasn't been checked in.


yattias (approved):
LGTM

PR #1468, opened 2024-03-18
Remove browsable API from production, some features are exposing user data

koutst:
- We currently have the browsable API in prod. Some of the views have a dropdown of all users. We should disable in production (and staging?).
- What is happening?
  - A user has a foreign key which due to the way the serializer is set up the Django framework automatically shows in the form view as a dropdown. The result is a dropdown of all our users.

### DIFF
--- a/src/researchhub/settings.py
+++ b/src/researchhub/settings.py
@@ -310,6 +310,13 @@ FILE_UPLOAD_MAX_MEMORY_SIZE = 26214400 * 24.1  # ~655MB max data allowed
 
 PAGINATION_PAGE_SIZE = 10
 
+DEFAULT_RENDERER_CLASSES = ("rest_framework.renderers.JSONRenderer",)
+
+if DEBUG:
+    DEFAULT_RENDERER_CLASSES = DEFAULT_RENDERER_CLASSES + (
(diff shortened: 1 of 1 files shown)


src/researchhub/settings.py
```diff
@@ -310,6 +310,13 @@ def silky_capture(request):
 
 PAGINATION_PAGE_SIZE = 10
 
+DEFAULT_RENDERER_CLASSES = ("rest_framework.renderers.JSONRenderer",)
+
+if DEBUG:
+    DEFAULT_RENDERER_CLASSES = DEFAULT_RENDERER_CLASSES + (
+        "rest_framework.renderers.BrowsableAPIRenderer",
+    )
+
```
gzurowski, line 319: I would remove this as well, so the application behaves the same across any environment.

gzurowski (approved):
LGTM 👍

src/researchhub/settings.py
koutst: I didn't want to disrupt Kobe's workflow if he uses it but I agree that all environments should be consistent. @yattias  thoughts?

yattias (approved):
LGTM Taki

PR #1470, opened 2024-03-19
Dev Container | Copy static content after container creation

gzurowski:
After container creation, copy static assets to avoid issues resolving includes in template files and similar issues.


yattias (approved):
LGTM

koutst (approved):
LGTM

PR #1472, opened 2024-03-19
Add claim author check

gzurowski:
When claiming authorship of a paper, verify that regular users cannot claim for others than themselves.


koutst (approved):
LGTM

yattias (approved):
This looks good and we should merge this. I wonder if we should create a backlog task to update `AuthorClaimCaseSerializer` to extract `requestor` from the auth token instead of examining the id

gzurowski:
> This looks good and we should merge this. I wonder if we should create a backlog task to update `AuthorClaimCaseSerializer` to extract `requestor` from the auth token instead of examining the id

I will take a quick look tomorrow, but let's create a backlog item for this.

yattias (approved):
LGTM

PR #1473, opened 2024-03-19
Fix org query set

koutst:
- The Organizations queryset wasn't filtering Organizations by user so a user could get access to all organizations.

### DIFF
--- a/src/user/views/organization_view.py
+++ b/src/user/views/organization_view.py
@@ -39,6 +39,13 @@ class OrganizationViewSet(viewsets.ModelViewSet):
         "slug",
     ]
 
+    def get_queryset(self):
+        user = self.request.user
+        organizations = self.queryset.filter(
+            permissions__user=user,
+        ).distinct()
+        return organizations
+
     def get_object(self, slug=False):
         queryset = self.filter_queryset(self.get_queryset())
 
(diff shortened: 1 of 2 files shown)


src/user/views/organization_view.py
```diff
@@ -124,7 +131,10 @@ def _create_image_file(self, data, organization, user):
     def update(self, request, *args, **kwargs):
         user = request.user
         partial = kwargs.pop("partial", False)
-        organization = self.get_object()
+        try:
+            organization = self.get_object()
+        except Exception:
+            return Response({"data": "No permission to get organization"}, status=403)
```
gzurowski, line 137: I am wondering if we can assume an access issue for any type of exception?

gzurowski (approved):
LGTM 👍

src/user/views/organization_view.py
yattias: I think so

yattias (approved):
This looks good

src/user/views/organization_view.py
koutst: `get_object` returns an Http404 error if not found so I will just check for that.

gzurowski: The approach can mask legitimate _not found_ cases, i.e., where a non-existing org is requested. I believe a HTTP 404 response for both cases should be fine.

PR #1476, opened 2024-03-20
Housekeeping | Pin service images in GitHub workflow

gzurowski:
Pin service images to the same tag that is used for the dev container setup, aligning with the version used across all environments.


koutst (approved):
LGTM

PR #1479, opened 2024-03-20
Add missing import statement to readme

kerkelae:


gzurowski:
As discussed over Slack, there is an issue running the database seeding snippet for creating hubs when using `python manage.py shell` (instead of `python manage.py shell_plus` as specified in the README):

```
NameError: name 'Hub' is not defined
```

I believe it's a good idea to include this patch, so it works in both types of shell.

gzurowski (approved):
LGTM 👍

Thanks for your contribution @kerkelae!

PR #1480, opened 2024-03-20
Filter purchase query on user querying

koutst:
- Any user can see the list of all purchases on the site. The list of purchases should be specific to the user.


gzurowski (approved):
LGTM 👍

yattias (approved):
LGTM

PR #1483, opened 2024-03-21
Filter deposits on requesting user

gzurowski:
Users querying the `/deposit` endpoint should only see their own deposits.
Staff users will be able to query all deposits, aligning with the withdrawal implementation in the same file.

### DIFF
--- a/src/reputation/tests/test_views.py
+++ b/src/reputation/tests/test_views.py
@@ -9,7 +9,7 @@ from rest_framework.test import APITestCase
 from reputation.distributions import Distribution as Dist
 from reputation.distributor import Distributor
 from reputation.models import Withdrawal
-from reputation.tests.helpers import create_withdrawals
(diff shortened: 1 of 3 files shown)


src/reputation/tests/test_views.py
```diff
@@ -22,6 +22,45 @@ def setUp(self):
...
+
+        create_deposit(user)
+
+        self.client.force_authenticate(other_user)
+        response = self.client.get("/api/deposit/")
+
+        self.assertEquals(response.status_code, 200)
+        self.assertEquals(response.data["count"], 0)
+
+    def test_deposit_deposit_user_can_list_all_deposits(self):
```
koutst, line 48: Should this test name reference that the user is a staff user?

koutst (approved):
LGTM

yattias (approved):

PR #1491, opened 2024-03-22
Only return sub in linkedin_data

koutst:
- linkedin_data currently returns email and other identifying information that isn't being used by the front end. The only field that looks to be used is the sub field.


yattias (approved):
LGTM

gzurowski (approved):
LGTM 👍

PR #1495, opened 2024-03-24
Dev Container | Disable auto forwarding ports

gzurowski:
Use the ports defined in `forwardPorts` instead of automatically discovering port (from processes or even parsed output from the integrated terminal).


koutst (approved):
LGTM

PR #1504, opened 2024-03-26
Housekeeping | Fix note tests

gzurowski:
- None of the note tests were running due to the tests package being non-discoverable.
- Fix existing tests and update assertions to align with the current implementation.

### DIFF
--- a/src/note/tests/test_note_api.py
+++ b/src/note/tests/test_note_api.py
@@ -168,13 +168,17 @@ class NoteTests(APITestCase):
         # Update body
         response = self.client.post(
             '/api/note_content/',
-            {'full_src': 'updated body', 'note': note['id'], 'plain_text': ''}
+            {
+                'full_src': 'updated body',
+                'note': note['id'],
+                'plain_text': 'updated body',
+            }
         )
 
         # Re-fetch note
         response = self.client.get(f"/api/note/{note['id']}/")
         note = response.data
-        self.assertEqual(note['latest_version']['src'], 'updated body')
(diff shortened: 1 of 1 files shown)


yattias (approved):
LGTM

src/note/tests/test_note_api.py
```diff
@@ -86,7 +86,11 @@ def test_cannot_create_shared_note_manually(self):
         # NOTE: Should only be able to created SHARED note by inviting useres
         self.assertNotEqual(created_note['access'], 'SHARED')
 
-    def test_note_editor_cannot_invite_others(self):
+    def test_note_editor_can_invite_others(self):
```
koutst, line 89: @yattias is a note editor supposed to be able to invite others? The old test name suggests they shouldn't be able to.

koutst (approved):
LGTM

src/note/tests/test_note_api.py
gzurowski: I am not familiar with the business rules around this. I renamed the test to reflect the behavior of the current implementation. It allows for this via the `IsOrganizationUser` permission class on the `invite_user` function.

koutst: Sorry, this was out of scope of this PR. I was just wondering if we should file a bug to fix later.

gzurowski:
@yattias I am merging this, but it would be good to get your input on https://github.com/ResearchHub/researchhub-backend/pull/1504#discussion_r1541007817.

PR #1511, opened 2024-03-28
Fix Organization Check

gzurowski:
- Don't allow `POST` requests to satisfy organization check.
- Add unit tests


koutst:
The only endpoint this looks like it might effect is the invite_user method in the notes view. Is it possible this will cause a similar issue to what we saw earlier this week where a user has access to a note but not the org and so they won't be able to invite other users when they should be able to?

PR #1512, opened 2024-03-28
Fix note permissions

gzurowski:
Return notes from the `/api/note` that requesting user has explicitly access for.


koutst (approved):
LGTM

PR #1514, opened 2024-03-29
Default to PENDING status to avoid wrongly marking transaction as FAILED

koutst:
- This might require manual correction but better than allowing double withdrawal.
- This is a temporary solution the actual solution here is to use open zeplin to receive webhooks when transactions land.


yattias (approved):
LGTM!

PR #1515, opened 2024-04-01
Use ElasticSearch version 7.10.1

gzurowski:
This change is mainly to align the ElasticSearch version we use across environments. This change updates the version in the dev container as well as in the Docker Compose and native installation setups.


yattias (approved):
LGTM

PR #1521, opened 2024-04-08
[Search] More fields in ES; Boost some fields;

yattias:
Adding a few more fields to elastic search and boosting the weight of some fields for the purpose of affecting the sort order

### DIFF
--- a/src/search/documents/paper.py
+++ b/src/search/documents/paper.py
@@ -161,7 +161,7 @@ class PaperDocument(BaseDocument):
     def prepare_can_display_pdf_license(self, instance):
         try:
             return pdf_copyright_allows_display(instance)
-        except:
+        except Exception:
             pass
 
         return False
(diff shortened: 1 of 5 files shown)


src/search/views/paper.py
```diff
@@ -41,7 +41,7 @@ def _is_doi(search_term):
     pagination_class = LimitOffsetPagination
     lookup_field = "id"
     # This field will be added to the ES _score
-    score_field = "score"
+    # score_field = "score"
```
gzurowski, line 44: Nit: Can we remove this instead of commenting it out?

src/search/documents/paper.py
```diff
@@ -134,16 +139,38 @@ def prepare_suggestion_phrases(self, instance):
...
+
     def prepare_paper_publish_year(self, instance):
         if instance.paper_publish_date:
             return instance.paper_publish_date.year
         return None
 
+    def prepare_can_display_pdf_license(self, instance):
+        try:
+            return pdf_copyright_allows_display(instance)
+        except:
```
gzurowski, line 164: Not: Avoid bare `except`s:

```suggestion
        except Exception:
```

gzurowski (approved):
LGTM 👍

PR #1524, opened 2024-04-12
Fix flaky withdraw test, hitting etherscan rate limit, add request mocking

koutst:


gzurowski (approved):
It would be good to keep the `requrements.txt` file in sync with the Poetry changes.

Otherwise, LGTM 👍

PR #1525, opened 2024-04-12
Add CoinGecko api key

koutst:
- We currently don't use an api key when making requests to coingecko. This limits us to 5 req/min. Adding an api key increases that to 30/min (I think that value is actually way higher). This will prevent issues like https://researchhub.sentry.io/issues/5070443073/?environment=prod&environment=production&project=1797024&query=is%3Aunresolved+withdraw&referrer=issue-stream&statsPeriod=90d&stream_index=12&utc=true


gzurowski (approved):
LGTM 👍

PR #1527, opened 2024-04-15
Do not call async service to update deposit status, rely on celery tasks

koutst:
- No longer call the async service to update the status of a deposit. Rely on celery tasks run every 5 minutes.
- The async service interacts with the db directly to update the deposit status to PAID and then calls the API to finalize the update. This is very unreliable.


yattias (approved):
LGTM

PR #1530, opened 2024-04-16
Fix withdrawal tests

koutst:
- Start actually testing if a withdrawal was successful. 
- We have to mock complete_token_transfer because EthereumTesterProvider is missing functionality.

### DIFF
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -21,7 +21,7 @@ django-extensions = "3.0.9"
 smart-open = "5.2.1"
 pre-commit = "2.17.0"
 django-ses = "3.5.2"
-web3 = {extras = ["tester"], version = "^6.17.1"}
+web3 = {version = "6.17.1", extras = ["tester"]}
 django-silk = "5.0.2"
 django-dbbackup = "4.0.2"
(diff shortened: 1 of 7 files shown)


src/pyproject.toml
```diff
@@ -21,7 +21,7 @@ django-extensions = "3.0.9"
 smart-open = "5.2.1"
 pre-commit = "2.17.0"
 django-ses = "3.5.2"
-web3 = "6.15.1"
+web3 = {extras = ["tester"], version = "^6.17.1"}
```
gzurowski, line 24: I suggest pinning this to a fixed version.

src/user/tests/helpers.py
```diff
@@ -111,6 +110,15 @@ def create_random_authenticated_user_with_reputation(unique_value, reputation):
     return user
 
 
+def create_old_random_authenticated_user_with_reputation(unique_value, reputation):
```
gzurowski, line 113: Nit: The name can be slightly confusing. Is it worth having an extra helper for this when it's only used once?

gzurowski (approved):
LGTM 👍

PR #1534, opened 2024-04-17
Lock user updates while tipping to prevent double spend

koutst:


yattias (approved):
I think this looks good.

gzurowski (approved):
LGTM 👍

PR #1540, opened 2024-04-19
[Cleanup] Remove external email addresses

gzurowski:


src/researchhub/settings.py
```diff
@@ -555,19 +555,13 @@ def silky_capture(request):
...
 EMAIL_WHITELIST = [
-    "leosun@quantfive.org",
-    "patrick@quantfive.org",
-    "patrick.lu@berkeley.edu",
     "pdj7@georgetown.edu",
     "bank@researchhub.com",
     "kobeattias@gmail.com",
     "kobe@researchhub.com",
     "kobe+1@researchhub.com",
     "contact@notesalong.com",
```
koutst, line 563: Should we remove this one too?

koutst (approved):
LGTM

src/researchhub/settings.py
gzurowski: I believe this is @yattias, not sure if it's still used?

I suggest to eventually move all of this into configuration, but maybe in a next step.

PR #1541, opened 2024-04-19
Cleanup | Remove distribution endpoint

gzurowski:
- Remove referral distribution endpoint and configuration.
- Remove inactive referral distribution code.

**Note:** As a follow up, we might want to remove the `referral` app altogether if that is no longer needed.


yattias (approved):
LGTM

PR #1544, opened 2024-04-22
Housekeeping | Remove Referral Django Application (Part 1)

gzurowski:
- Remove the `referral` Django application that is no longer used.
- Remove referral settings.
- Remove database objects with a corresponding database migration.

*Note*

A follow up PR will completely remove the application. This is to make sure the database migration (which is located within the `referral` application) runs in all environments.


gzurowski:
> LGTM. Only question I have is should we be deleting the tables? Is that information useful at all? Might we want to reinstate the referral program at some point?

As discussed with @yattias, we can remove the referral program as there are no plans to reinstantiate it. There are only 95 rows in the referral table, all of which expired last year.

PR #1548, opened 2024-04-23
Housekeeping | Remove async service

gzurowski:
The existing code related to async service is not used in any active code path and can be removed. Checking transaction status is being done with a Celery task.


yattias (approved):
Nice. happy to get rid of this

PR #1549, opened 2024-04-23
Add client id field to purchase

koutst:
- Check if client ID already exists in db before creating new purchase.


src/purchase/views/purchase_view.py
```diff
@@ -79,6 +79,20 @@ def create(self, request):
         with transaction.atomic():
             user = User.objects.select_for_update().get(id=user.id)
 
+            # Check if a purchase was already created for this client ID. If a purchase exists
+            # the user probably accidentally double clicked submit and we should return an error.
+            purchase = Purchase.objects.filter(
```
yattias, line 84: There may be some additional FE logic that does not pass a `client_id`. I wonder if it will throw an error. I think this is okay - I can update all spots to ensure `client_id` is passed

yattias (approved):

gzurowski (commented):
I would avoid mixing infrastructure-related semantics (duplicate requests) with business-related fields in the `purchase` table. We already use Redis as a distributed cache that is often used as a repository for creating idempotent receivers.

yattias:
@gzurowski Do you think passing in a unique `client_id` from the frontend is not needed?
Do you have other thoughts / ideas here? The big idea was to tie a transaction initiated from the FE to the BE.

koutst:
Yeah, this makes sense. I will close this PR and look into using Redis instead.

PR #1551, opened 2024-04-25
Add caching on client id for purchases

koutst:
- To prevent accidental double purchases such as tipping.
- All requests to this endpoint will require a client id.

### DIFF
--- a/src/purchase/views/purchase_view.py
+++ b/src/purchase/views/purchase_view.py
@@ -212,10 +212,10 @@ class PurchaseViewSet(viewsets.ModelViewSet):
                 )
                 distributor.distribute()
 
-        serializer = self.serializer_class(purchase, context=context)
-        serializer_data = serializer.data
+            serializer = self.serializer_class(purchase, context=context)
+            serializer_data = serializer.data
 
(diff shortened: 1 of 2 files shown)


yattias (approved):
LGTM

src/purchase/views/purchase_view.py
```diff
@@ -77,6 +78,17 @@ def create(self, request):
...
+            cached_client_id = cache.get(f"purchase_client_id_{client_id}")
+            if cached_client_id:
+                return Response(
+                    {
+                        "detail": "This purchase has already been processed",
+                    },
+                    status=400,
+                )
+
+            cache.set(f"purchase_client_id_{client_id}", True, timeout=60)
```
gzurowski, line 90: Nit: It might be better to implement idempotent behavior and return the original transaction with a 2xx series response code.

gzurowski (approved):
LGTM 👍

gzurowski (approved):
LGTM 👍

yattias (approved):
LGTM

koutst:
Paper contributions front end needs to be updated before deploying

PR #1554, opened 2024-04-30
Add validation to deposits

koutst:
- Validate deposit amount and to address.

### DIFF
--- a/src/reputation/tasks.py
+++ b/src/reputation/tasks.py
@@ -34,7 +34,7 @@ from utils.sentry import log_error, log_info
 
 DEFAULT_REWARD = 1000000
 
-TRANSACTION_AGE_LIMIT = 60 * 60 * 6  # 6 hours
+PENDING_TRANSACTION_TTL = 60 * 60 * 1  # 1 hour
 
 
 @app.task(queue=QUEUE_CONTRIBUTIONS)
@@ -158,7 +158,7 @@ def evaluate_transaction(transaction_hash):
     is_correct_to_address = func_params["to"] == WEB3_KEYSTORE_ADDRESS
     block_timestamp = datetime.fromtimestamp(block["timestamp"])
     is_recent_transaction = block_timestamp > datetime.now() - timedelta(
-        seconds=TRANSACTION_AGE_LIMIT
+        seconds=PENDING_TRANSACTION_TTL
     )
 
     return (
@@ -179,9 +179,9 @@ def check_deposits():
     deposits = Deposit.objects.filter(paid_status=None).order_by("created_date")
 
     for deposit in deposits.iterator():
-        # If a deposit is not resolved after 6 hours, mark it as failed
+        # If a deposit is not resolved after our set TTL, mark it as failed
         if deposit.created_date < datetime.now(pytz.UTC) - timedelta(
-            seconds=TRANSACTION_AGE_LIMIT
+            seconds=PENDING_TRANSACTION_TTL
         ):
             deposit.set_paid_failed()
             continue
@@ -214,7 +214,7 @@ def check_deposits():
                 deposit.amount = deposit_amount
                 deposit.set_paid()
         except Exception as e:
-            log_error(e)
+            log_error(e, "Failed to process deposit")
             deposit.set_paid_pending()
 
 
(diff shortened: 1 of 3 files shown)


src/reputation/tasks.py
```diff
@@ -31,6 +34,8 @@
 
 DEFAULT_REWARD = 1000000
 
+TRANSACTION_AGE_LIMIT = 60 * 60 * 6  # 6 hours
```
yattias, line 37: nit^2: Can we change to something like `RECENT_TRANSACTION_LIMIT` or `RECENT_TRANSACTION_TTL` ? I wasn't sure what this meant without looking at it in context

src/reputation/tasks.py
```diff
@@ -103,21 +108,63 @@ def create_author_contribution(contribution_type, user_id, unified_doc_id, objec
...
-        log_error(e)
-        print(e)
-        return False
+
+def get_block(block_number):
+    # Check if connected successfully
+    if not w3.is_connected():
+        raise Exception("Failed to connect to Ethereum node.")
+
+    return w3.eth.get_block(block_number)
```
yattias, line 137: Should we check for a scenario where block number does not exist? I'm not sure if it is possible. Not sure what  would happen in such scenario

src/reputation/tasks.py
```diff
@@ -126,16 +173,49 @@ def check_transaction_success(transaction_hash):
...
 def check_deposits():
-    deposits = Deposit.objects.filter(paid_status=None)
+    # Sort by created date to ensure a malicious user doesn't attempt to take credit for
+    # a deposit made by another user. This is a temporary solution until we add signed messages
+    # to validate users own wallets.
+    deposits = Deposit.objects.filter(paid_status=None).order_by("created_date")
+
     for deposit in deposits.iterator():
-        amt = deposit.amount
+        # If a deposit is not resolved after 6 hours, mark it as failed
```
yattias, line 182: I wonder if we need to reduce the time here. 6 hours seems like a lot?

src/reputation/tasks.py
```diff
@@ -126,16 +173,49 @@ def check_transaction_success(transaction_hash):
...
+        if deposit.created_date < datetime.now(pytz.UTC) - timedelta(
+            seconds=TRANSACTION_AGE_LIMIT
+        ):
+            deposit.set_paid_failed()
+            continue
+
+        deposit_previously_paid = Deposit.objects.filter(
```
yattias, line 189: So these would be deposits with the same transaction hash? Does it make sense to require that column to be unique so we don't get duplicates as such in the first place?

yattias (approved):
Had a few questions but LGTM

src/reputation/tasks.py
```diff
@@ -126,16 +173,49 @@ def check_transaction_success(transaction_hash):
...
+                continue
+
+            with transaction.atomic():
+                distribution = Dist("DEPOSIT", deposit_amount, give_rep=False)
+                distributor = Distributor(distribution, user, user, time.time(), user)
+                distributor.distribute()
+                deposit.amount = deposit_amount
+                deposit.set_paid()
+        except Exception as e:
+            log_error(e)
```
gzurowski, line 217: Nit: Should we add a `message` here for more context?

gzurowski (approved):
LGTM 👍

src/reputation/tasks.py
koutst: I don't think this is possible since we check the transaction receipt was successful which I believe would require a block number.

src/reputation/tasks.py
koutst: Yeah, that's what I was originally thinking but the db seems to have duplicates and I was trying to find the easiest way to get this up and running. Happy to clean up the db and add uniqueness constraints.

src/reputation/tasks.py
koutst: I will call it `PENDING_TRANSACTION_TTL`. What do you think?

src/reputation/tasks.py
koutst: Set to 1 hour for now.

src/reputation/tasks.py
yattias: Sounds good

src/reputation/tasks.py
yattias: Yeah 1 hour seems good

PR #1556, opened 2024-05-01
Housekeeping | Remove API key for async service

gzurowski:
The async service is no longer used (and is shut down), therefore this change removes the following:
- The API key configuration settings.
- A permission class using the API key.
- An API method that had been used by async service for depositing RSC.


koutst (approved):
LGTM

yattias (approved):
Nice!

PR #1559, opened 2024-05-03
[Institutions] Load institutions from OpenAlex

yattias:
- Institution model
- Script to upsert institutions from OpenAlex


koutst (approved):
LGTM

PR #1560, opened 2024-05-03
Create new provider url and hotwallet address variables, update hotwallet address

koutst:
- [ ] Transfer funds right before merge to limit low balance alerts.


yattias (approved):
LGTM

PR #1566, opened 2024-05-06
add web3 network placeholder to local env

thetechnocrat-dev:
The Dev containers setup and DB seed worked well 💯 . I just had to add one env variable to the sample file that was missing. 

This is the error that this change fixes:

Screenshot 2024-05-06 at 4 37 58 PM


gzurowski (approved):
LGTM 👍 

Thanks for your contribution.

PR #1568, opened 2024-05-07
1 upvote equals 1 RSC

koutst:
- Remove enhanced upvote functionality.

### DIFF
--- a/src/reputation/tests/test_upvote_rsc_distribution.py
+++ b/src/reputation/tests/test_upvote_rsc_distribution.py
@@ -4,12 +4,9 @@ import math
 from django.contrib.admin.options import get_content_type_for_model
 from django.test import TestCase
 
+import reputation.distributions as distributions
 from discussion.models import Vote as GrmVote
 from discussion.tests.helpers import create_rh_comment
-from reputation.distributions import (
-    calculate_rsc_per_upvote,
-    create_upvote_distribution,
-)
 from reputation.models import Distribution, Escrow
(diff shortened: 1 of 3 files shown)


src/reputation/tests/test_upvote_rsc_distribution.py
```diff
@@ -219,63 +127,5 @@ def test_ineligible_enhanced_distribution(self):
...
         eligible_user.save()
 
-        distribution_amount = calculate_rsc_per_upvote()
-        distribution = create_upvote_distribution(
-            1, self.original_paper, GrmVote.objects.first()
-        )
+        distribution = distributions.Distribution(1, 1, 1)
         self.assertEquals(distribution.amount, 1)
-
-    def test_author_claim_pot(
```
yattias, line 228: The claim tests might be different than this feature.

yattias (approved):
LGTM but one comment

src/reputation/tests/test_upvote_rsc_distribution.py
koutst: It seems like it's testing the escrow payout for escrow that used to be created in the create_upvote_distribution method. Since that method is no longer used and the escrow isn't created, this isn't really testing anything anymore.

koutst: Synced with Kobe. I will remove as this is not really testing anything anymore.

PR #1592, opened 2024-05-14
Load topics from OpenAlex + Refactor of OpenAlex work processor

yattias:
- Refactor `process_openalex_works` which is responsible for upserting works and their related entities by eliminating dead code and removing unnecessary code that significantly slows down paper processing such as `hydrate_paper_concepts`
- Added a django script to fetch and create topics from OpenAlex and their related  entities such as `subfield`, `field`, and `domain`
- Added a script `load_works_from_openalex` which has two modes: `fetch` or `backfill`. This script will be iterated upon as we will need to do a bunch of additional operations such as creating authors when fetching papers 
- Enriched paper entity by adding various new fields such as `work_type`, `is_retracted`, ....


src/paper/openalex_util.py
```diff
@@ -0,0 +1,146 @@
+import logging
```
yattias, line 1: The code below is a refactored + optimized version of the code below:

https://github.com/ResearchHub/researchhub-backend/pull/1592/files#diff-28e810c7045544f9f4b257545fa75c388004e8b44fbbdb822e71eb8886860746L914

yattias: It is 50% existing + 50% new I'd say

src/topic/management/commands/load_topics_from_openalex.py
```diff
@@ -0,0 +1,38 @@
+from django.core.management.base import BaseCommand
+
+from topic.models import Topic
+from utils.openalex import OpenAlex
+
+
+class Command(BaseCommand):
```
gzurowski, line 7: Since were are starting a new app with `topic`, it would be good to have some tests for this command?

gzurowski (approved):
LGTM 👍 

The changed code and the newly introduced code could use some tests. :)

src/topic/models.py
```diff
@@ -0,0 +1,244 @@
...
+
+        topic = None
+        try:
+            topic = Topic.objects.get(openalex_id=oa_topic["id"])
+        except Topic.DoesNotExist:
+            pass
+
+        # if topic exists, determine if we need to update it
+        needs_update = False
+        if topic and has_dates:
```
koutst, line 144: Why might a topic not have dates? Are there scenarios where we would want to update topics that don't have dates?

koutst (approved):
LGTM

src/topic/management/commands/load_topics_from_openalex.py
yattias: Yes, I was planning on adding some tests today!

PR #1602, opened 2024-05-17
[Author] 2 yr mean citedness; author institution

yattias:
- Add `AuthorInstitution` model
- Add "impact factor" also known as 2 year mean citedness
- Remove `AuthorCitation` model


src/paper/openalex_util.py
```diff
@@ -156,3 +159,40 @@ def process_openalex_works(works):
     for paper_id, topics_and_concepts in paper_to_topics_and_concepts.items():
         openalex_concepts, openalex_topics = topics_and_concepts
         create_paper_related_tags(paper_id, openalex_concepts, openalex_topics)
+
+    openalex_authorships = work.get("authorships")
+
+    print("work:", work)
```
gzurowski, line 165: Is that a leftover?

src/user/related_models/author_model.py
```diff
@@ -9,6 +9,7 @@
 from paper.utils import PAPER_SCORE_Q_ANNOTATION
 from purchase.related_models.purchase_model import Purchase
 from researchhub_case.constants.case_constants import APPROVED
+from user.related_models.author_institution import AuthorInstitution
```
gzurowski, line 12: Nit: Is that change needed?

gzurowski (approved):
LGTM 👍

src/user/related_models/author_model.py
yattias: Yes, it is needed because there is no `_init.py` file in that directory which includes the model. Django is unable to "see" it. I could alternatively create an _init file but didn't feel it was needed

PR #1603, opened 2024-05-21
Create Authors, co-authors as we process papers from OpenAlex

yattias:
As we process papers from OpenAlex, create authors associated with the paper along with related metadata which includes:
- Stats (e.g. `h_index`, `i10_index`)
- Coauthors
- Author Institutions


src/paper/openalex_util.py
```diff
@@ -156,18 +158,42 @@ def process_openalex_works(works):
             sentry.log_error(e, message="Failed to bulk update papers")
 
     # Upsert concepts and associate to papers
-    for paper_id, topics_and_concepts in paper_to_topics_and_concepts.items():
-        openalex_concepts, openalex_topics = topics_and_concepts
-        create_paper_related_tags(paper_id, openalex_concepts, openalex_topics)
+    for paper_id, paper_data in paper_to_openalex_data.items():
+        print("processing work: " + paper_data["openalex_work"]["id"])
```
gzurowski, line 162: I am wondering if we want these type of side-effects in a utility function, and if so, whether we should use an actual logger that includes some contextual metadata (log level, timestamp, formatting at a minimum) . It's fully understood that `print`s have spread across the code base, but I suggest we stop this practice.

src/paper/openalex_util.py
```diff
@@ -181,18 +207,100 @@ def process_openalex_works(works):
...
+                    source=AuthorContributionSummary.SOURCE_OPENALEX,
+                    author=author,
+                    year=activity.get("year"),
+                    defaults={
+                        "works_count": activity.get("works_count", None),
+                        "citation_count": activity.get("cited_by_count", None),
+                    },
+                )
+            except Exception as e:
+                sentry.log_error(
```
gzurowski, line 277: Should this be logged as well?

src/paper/paper_upload_tasks.py
```diff
@@ -707,13 +707,17 @@ def create_paper_related_tags(paper_id, openalex_concepts=[], openalex_topics=[]
...
-            except Exception as e:
-                sentry.log_error(
-                    e, message=f"Failed to add paper to biorXiv hub: {paper.id}"
-                )
+            if Hub.objects.filter(id=biorxiv_hub_id).exists():
+                try:
+                    paper.hubs.add(biorxiv_hub_id)
+                    paper.unified_document.hubs.add(biorxiv_hub_id)
+                except Exception as e:
+                    sentry.log_error(
```
gzurowski, line 715: Should this be logged as well?

gzurowski (approved):
LGTM 👍

src/paper/openalex_util.py
yattias: Let's talk about a better way to do this during our sync

PR #1604, opened 2024-05-22
[Profile] Endpoint for fetching profile data

yattias:
- Endpoint for fetching profile data for new profile page
- Individual contribution cards (comments, papers, etc..) will continue to operate as they do now

🧪  Tests to follow before merging this

- [ ] Add tests


src/user/views/author_views.py
```diff
@@ -0,0 +1,976 @@
...
+
+        if getattr(instance, "_prefetched_objects_cache", None):
+            # If 'prefetch_related' has been applied to a queryset, we need to
+            # forcibly invalidate the prefetch cache on the instance.
+            instance._prefetched_objects_cache = {}
+
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["get"], permission_classes=[AllowAny])
+    def profile(self, request, pk=None):
```
yattias, line 122: This method is the only change here. This contents of this file was extracted from `user_views.py`

PR #1616, opened 2024-06-03
Accept `ALLOWED_HOSTS` from environment

gzurowski:
Accept additional hosts via the environment variable `ALLOWED_HOSTS`. This will be used to e.g. add the Beanstalk load balancers after provisioning.


yattias (approved):
Good idea

koutst (approved):

PR #1617, opened 2024-06-04
Paper revisions

yattias:
- Keep track of paper revisions
- Helps us figure out the number of citations changed since last time, open access status change, ...
- Can be used to display revisions of a paper on the client which is a future goal 

Closes https://github.com/ResearchHub/issues/issues/10


src/paper/migrations/0127_historicalpaper.py
yattias, line 1: This migration is auto-generated by the simple history package

src/paper/related_models/paper_model.py
```diff
@@ -59,6 +60,7 @@
 
 
 class Paper(AbstractGenericReactionModel):
+    history = HistoricalRecords()
```
yattias, line 63: This is how we tell the package to track history

koutst (approved):
LGTM

src/pyproject.toml
```diff
@@ -83,6 +83,7 @@ urllib3 = "1.26.18"
 aiohttp = "3.9.5"
 cryptography = "42.0.5"
 requests-mock = "^1.12.1"
+django-simple-history = "^3.7.0"
```
gzurowski, line 86: I suggest to target a specific version instead of a range.

yattias: Good catch, I thought I did

PR #1618, opened 2024-06-04
Migrating to Django 4.2 LTS

yattias:
- Migrating django to an LTS version 
- 4.2 is required by the latest version of the simple-history package #1617


koutst (approved):

gzurowski (approved):
LGTM 👍

PR #1630, opened 2024-06-12
Proposed model for rep v2

koutst:
Proposal for rep v2.
- Considerations:
  - Auditable
    - We want to be able to understand exactly how we came to a calculation at any point in time. 
  - History page
    - Display to users how and why rep changed over time.
  - Versioning
    - We want to be able to update the algorithm and recalculate the reputation history and current reputation.
    - Bins will vary over time and need to be updated and scores need to be recalculated.
  
  
- Some example scenarios https://docs.google.com/spreadsheets/d/1tlGB-kBugXzreudA8qswDytmRbhXigSI_7thT7eeACU/edit#gid=0


yattias (approved):
LGTM

PR #1634, opened 2024-06-16
[Personalize] Updating filter and campaign names

yattias:
Updating filter and campaign names in prep for new env change


gzurowski (approved):
LGTM 👍

PR #1635, opened 2024-06-17
Mock rep v2 in author profile

koutst:


yattias (approved):

src/user/serializers.py
```diff
@@ -1050,6 +1050,7 @@ def get_coauthor(self, coauthor):
 class DynamicAuthorProfileSerializer(DynamicModelFieldSerializer):
     institutions = SerializerMethodField()
     coauthors = SerializerMethodField()
+    reputations = SerializerMethodField()
```
yattias, line 1053: Could we change name to `reputation_list` or a different name? I'm thinking we will also need a field called `reputation` to indicate what the user's rep is (atm it is decided it will be the highest value in that list: `max_val(reputations)`

src/user/serializers.py
```diff
@@ -1129,7 +1130,16 @@ def get_open_access_pct(self, author):
             / total_paper_count
         )
 
-    def get_reputations(self, author):
+    def get_reputation(self, author):
+        return {
+            "hub_id": 1,
+            "hub_name": "Cell Biology",
+            "hub_slug": "slug-1",
+            "score": 1982000,
```
yattias, line 1138: This value is greater than the top bin value?

koutst: Will update the bins in the next PR.

koutst: Nvm, updated.

PR #1637, opened 2024-06-18
Associate hub with subfields

yattias:
- Add associated between hub and subfields
- This is a prerequisite for calculating reputation
- Includes script for creating subfield hubs
- Next step: Updating document hubs to include these newly created hubs


gzurowski (approved):
LGTM 👍

PR #1642, opened 2024-06-26
Script to tag existing papers with subfield hubs

yattias:


src/researchhub_document/management/commands/backfill_papers_from_openalex.py
```diff
@@ -0,0 +1,71 @@
...
+    def handle(self, *args, **kwargs):
+        OA = OpenAlex()
+        papers = Paper.objects.all()
+
+        if kwargs["id"]:
+            papers = papers.filter(id=kwargs["id"])
+        elif kwargs["start_id"]:
+            papers = papers.filter(id__gte=kwargs["start_id"])
+
+        if kwargs["interacted_content_only"]:
```
yattias, line 52: @koutst I think we don't need to retag every single paper in our DB. To speed things up, we can start with only papers that were interacted with (papers associated with comments + historically claimed)

gzurowski (approved):

PR #1652, opened 2024-07-03
Dev Containers | Align line length between black and flake

gzurowski:
Align line length between black and flake to 88 ([black's default](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length))


koutst (approved):
LGTM. Is 88 actually the right length?

gzurowski:
> LGTM. Is 88 actually the right length?

I don't think there is an agreed upon _right_ length in the Python community, but this is black's default: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length.

PR #1672, opened 2024-07-11
Fix rep_bins.csv path

koutst:


yattias (approved):

src/reputation/management/commands/create_rep_algo_vars.py
```diff
@@ -4,21 +4,23 @@
...
 import json
+import os
 
 from django.core.management.base import BaseCommand
 from django.db import transaction
 from django.db.models import Q
 
 from hub.models import Hub
 from reputation.models import AlgorithmVariables
+from researchhub.settings import BASE_DIR
```
gzurowski, line 15: Settings shouldn't be imported directly (see https://docs.djangoproject.com/en/5.0/topics/settings/#using-settings-in-python-code).

Instead, import with `from django.conf import settings` and get the setting like this: `settings.BASE_DIR`.

gzurowski (approved):
LGTM 👍

PR #1676, opened 2024-07-12
Process Persona webhook events

gzurowski:
- Add a new database model for storing user verification status.
- Add processing logic to the existing Person webhook handler to extract the relevant attributes from the event payload and store it in the database.


yattias (approved):

koutst (approved):
LGTM

PR #1682, opened 2024-07-15
Do not throw on missing author profile

koutst:
- There seems to be users that do not have author profiles. Skip calculation, do not throw.

### DIFF
--- a/src/reputation/management/commands/initialize_reputation.py
+++ b/src/reputation/management/commands/initialize_reputation.py
@@ -3,6 +3,7 @@ Calculate rep for a given author.
 """
 
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.management.base import BaseCommand
 from django.db import transaction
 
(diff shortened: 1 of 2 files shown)


src/reputation/management/commands/initialize_reputation.py
```diff
@@ -14,7 +15,10 @@
 
 def calculate_user_score(user_id, algorithm_version, recalculate=False):
     user = User.objects.get(id=user_id)
-    author = user.author_profile
+    try:
+        author = user.author_profile
+    except ObjectDoesNotExist:
+        return "User does not have an author profile."
```
gzurowski, line 21: This way, the function returns two different types, either a numeric type or a string. Handling of the exception could be done in the caller to retain a uniform return type.

gzurowski (approved):

src/reputation/management/commands/initialize_reputation.py
koutst: I will address this in a future PR

PR #1685, opened 2024-07-15
Return rep scores, remove mocks

koutst:
- Start returning rep scores in author profile

### DIFF
--- a/src/user/serializers.py
+++ b/src/user/serializers.py
@@ -23,7 +23,7 @@ from hypothesis.models import Hypothesis
 from institution.serializers import DynamicInstitutionSerializer
 from paper.models import Paper, PaperSubmission
 from purchase.models import Purchase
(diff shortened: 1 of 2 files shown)


src/user/serializers.py
```diff
@@ -1135,82 +1135,51 @@ def get_open_access_pct(self, author):
             )
 
     def get_reputation(self, author):
+        score = Score.objects.filter(author=author).order_by("-score").first()
+
+        if score is None:
+            return {}
```
koutst, line 1141: @yattias should I be returning an empty dict when score isn't calculated?

yattias: This is the primary hub, right?  Better to return None

yattias (approved):

gzurowski (approved):
LGTM 👍

PR #1686, opened 2024-07-16
Consolidating publication endpoints

yattias:
Combining `claim_profile_and_add_publications` and `add_publications` into a single endpoint

### DIFF
--- a/src/user/related_models/author_model.py
+++ b/src/user/related_models/author_model.py
@@ -343,18 +343,15 @@ class Author(models.Model):
                         elif vote.vote_type == 2:
                             vote_value = -1
 
-                        try:
-                            Score.update_score(
-                                self,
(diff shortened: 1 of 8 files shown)


src/user/related_models/author_model.py
```diff
@@ -407,12 +412,15 @@ def _calculate_score_hubs_citations(self, algorithm_version, score_version):
...
+                            self,
+                            hub,
+                            algorithm_version,
+                            score_version,
+                            citation_change,
+                            "citations",
+                            paper.id,
+                        )
+                    except Exception as e:
+                        print(e)
```
koutst, line 426: If we're going to add these `try`s we should add more context to prints.

koutst (approved):

PR #1691, opened 2024-07-16
Add reputation_v2 and reputation_list to user author_profile

koutst:
- Add reputation and reputation list to user author profile

### DIFF
--- a/src/user/serializers.py
+++ b/src/user/serializers.py
@@ -62,6 +62,8 @@ class AuthorSerializer(ModelSerializer):
     num_posts = SerializerMethodField()
     orcid_id = SerializerMethodField()
     reputation = SerializerMethodField()
+    reputation_v2 = SerializerMethodField()
+    reputation_list = SerializerMethodField()
     sift_link = SerializerMethodField()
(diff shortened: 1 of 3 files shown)


src/user/serializers.py
```diff
@@ -708,10 +759,14 @@ def _group_and_serialize_actions(self):
...
-                    data["paper_title"] = (
+                    data[
+                        "paper_title"
+                    ] = (
                         thread_post.title
                     )  # paper_title instead of post_title for symmetry on the FE
-                    data["paper_id"] = (
+                    data[
+                        "paper_id"
+                    ] = (
```
gzurowski, line 769: Weird, our formatters are off. I remember this was just changed by one of my previous commits...

gzurowski (approved):
LGTM 👍

src/user/serializers.py
koutst: My editor has it the original way but the pre commit changes it. Haven't looked into why.

gzurowski: Got it, we should take a look at that.

PR #1692, opened 2024-07-16
Add permission for checking verified users

gzurowski:
- Add a new permission class `IsVerifiedUser` for checking a user if KYC verified.
- Add the permission to the existing claiming endpoint.

**Note:**
The permission is not active in production (see code change). The conditional logic that is checking for `PRODUCTION` needs to be removed for releasing this feature.


koutst (approved):

gzurowski:
Closing in favor of #1694

PR #1719, opened 2024-07-25
[Paper Claim] Update feature to support reward use-case

yattias:
- Update Paper Claim flow to support new use cases
- Added `version` to claim records to differentiate old (deprecated) claims from new ones
- Placeholder for @koutst to add his code in
- Notify users after claim approved
- Stop the sending of emails in claim context


koutst (approved):
LGTM

PR #1723, opened 2024-07-26
Add operation for deleting authorships

gzurowski:
Add a new operation for deleting authorships (author-paper associations) on the `/author/<ID>/publications` resource using the HTTP `DELETE` method. The operation support deletion of multiple authorships by submitting multiple `paper_ids` in the request payload.


yattias (approved):
LGTM

PR #1730, opened 2024-07-27
Use new rep for determining if a user can withdraw

koutst:
- Re-enable skipped test.
- Add methods for getting scores.
- Remove unused oz webhook endpoint.
- Update check for rep score in withdrawal to use new rep score.

### DIFF
--- a/src/reputation/views/withdrawal_view.py
+++ b/src/reputation/views/withdrawal_view.py
@@ -61,61 +61,6 @@ class WithdrawalViewSet(viewsets.ModelViewSet):
         else:
             return Withdrawal.objects.filter(user=user)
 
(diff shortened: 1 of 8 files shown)


src/reputation/views/withdrawal_view.py
```diff
@@ -140,9 +85,9 @@ def create(self, request):
                 status=400,
             )
 
-        if user.reputation < 110:
+        if user.author_profile.get_rep_score() < 10:
```
yattias, line 88: @koutst we also need a condition for users that have their identity verified. So this should be:

`if not (sufficient_rep or user_is_verified): Error`

PR #1731, opened 2024-07-28
Use `is_verified` property over status query

gzurowski:
Use the existing `is_verified` property for checking if a user is verified.


yattias (approved):
LGTM

koutst (approved):

PR #1732, opened 2024-07-28
[Author Publications] Filter by author_id,paper_id pair

yattias:
- Update remove publication endpoint to filter by `author_id, paper_id` pair
- Add `work_type` to unified document serializer
- Remove `PROTECT` on author claim


gzurowski (approved):
LGTM 👍

PR #1738, opened 2024-07-29
Return RSC reward value

koutst:
- Replace mock data with real RSC reward value

### DIFF
--- a/src/paper/views/paper_views.py
+++ b/src/paper/views/paper_views.py
@@ -57,6 +57,11 @@ from paper.utils import (
 )
 from purchase.models import Purchase
 from reputation.models import Contribution
+from reputation.related_models.paper_reward import (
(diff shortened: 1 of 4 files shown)


src/paper/views/paper_views.py
```diff
@@ -583,9 +582,21 @@ def eligible_reward_summary(self, request, pk=None):
...
+            return Response("paper_id is required", status=400)
+
+        paper = Paper.objects.get(id=paper_id)
+        if not paper:
+            return Response("Paper not found", status=404)
+
+        try:
+            paper_rewards = paper.paper_rewards
+        except Exception:
+            return Response("Failed to get paper reward", status=400)
```
gzurowski, line 596: As we are past argument validation, we should return a server error (500 series).

gzurowski (approved):
LGTM 👍

PR #1746, opened 2024-07-30
DOI parsing in filtering function throws error sometimes

yattias:


src/utils/openalex.py
```diff
@@ -386,7 +386,8 @@ def get_works(
         # have a `researchhub` namespace in the DOI.
         filtered_works = list(
             filter(
-                lambda w: "/researchhub." not in w.get("doi", "").lower(),
+                lambda w: w.get("doi") is not None
```
gzurowski, line 389: This will filter out works without a DOI. The lambda should return work where DOI is `None` or where the namespace check passes.

PR #1754, opened 2024-07-30
User Authorship model in author overview

gzurowski:
Use the `Authorship` model for fetching documents in the author overview endpoint, replacing the deprecated `authored_papers` relation.


yattias (approved):
LGTM

PR #1768, opened 2024-08-02
Separate profile endpoint into 3 endpoints: profile, achievements, su…

yattias:
- Separate profile endpoint into 3 endpoints: profile, achievements, summary_stats
- Cache summary_stats, achievements for an hour to optimize (no purge mechanism, just TTL of one hour)

PR #1773, opened 2024-08-04
Small optimizations for summary stats method

gzurowski:
- Get user once at beginning of method
- Simplify query statements


yattias:
@gzurowski I did some refactor when splitting the endpoints. Can you update after I merge here https://github.com/ResearchHub/researchhub-backend/pull/1768

PR #1776, opened 2024-08-05
Reduce wallet withdrawal limit for verified users to 1 day, remove 14 day lock after account creation for verified users

koutst:


yattias (approved):
LGTM

PR #1781, opened 2024-08-07
Cache response of author `publications` and `overview` resources

gzurowski:
- Add caching to author `publications` and author `overview` resources.
- Update existing tests in `user` app to use test files from its own app (not cross-referencing files from the `paper` app).
- Move all JSON test files to `test_files` subdirectory.
- Invalidate author caches after adding/deleting publications

Closes https://github.com/ResearchHub/issues/issues/86


src/user/tests/test_views.py
```diff
@@ -42,9 +43,78 @@ def setUp(self):
...
+            f"author-{self.user_with_published_works.author_profile.id}-publications"
+        )
+        self.assertTrue(cache.get(cache_key))
+        self.assertEqual(cache.get(cache_key)[0]["documents"]["id"], paper.id)
+
+    def test_get_publications_reads_from_cache(self):
+        # Arrange
+        self.client.force_authenticate(self.user_with_published_works)
+
+        cached_data = ["cached"]
```
yattias, line 99: That's a nice way to test cached data

src/user/views/author_views.py
```diff
@@ -725,6 +732,8 @@ def publications(self, request, pk=None):
 
         serializer_data = serializer.data
 
+        cache.set(cache_key, serializer_data, timeout=3600)
+
         return self.get_paginated_response(serializer_data)
```
yattias, line 737: I wonder if we should cache the paginated response instead. Without the paginated response there will be no pagination if returned from cache

src/user/views/author_views.py
```diff
@@ -753,14 +762,12 @@ def add_publications(self, request, pk=None):
...
-                pull_openalex_author_works_batch.apply_async(
-                    (
-                        openalex_ids,
-                        request.user.id,
-                    ),
-                    priority=1,
-                )
+                chain(
+                    pull_openalex_author_works_batch.s(openalex_ids, request.user.id),
+                    invalidate_author_profile_caches.s(author.id),
```
yattias, line 769: This guarantees`invalidate_author_profile_caches` is fired after `pull_openalex_author_works_batch.` finishes?

yattias (commented):
Looks good. Had a comment and some questions

src/user/views/author_views.py
gzurowski: Thanks for pointing it out. I believe we shouldn't cache paginated results like this. Instead, I changed the code to cache the retrieved documents and wrap it in the pagination object accordingly.

yattias (approved):
LGTM

PR #1801, opened 2024-08-12
Adding DRF debugger

yattias:
- Debugger and profiler toolbar available when running api requests in the browser

Example
image

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,6 +10,10 @@
                 "runserver",
                 "[::]:8000"
             ],
+            // Uncomment this to view debugger in /api requests
+            // "env": {
+            //     "USE_DEBUG_TOOLBAR": "True",
+            // },
             "django": true,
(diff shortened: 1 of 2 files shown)


koutst (approved):

.vscode/launch.json
```diff
@@ -10,6 +10,10 @@
                 "runserver",
                 "[::]:8000"
             ],
+            // Uncomment this to view debugger in /api requests
+            // "env": {
+            //     "USE_DEBUG_TOOLBAR": "True",
+            // },
```
gzurowski, line 16: Since the launch configuration under version control, I would suggest either having this enabled by default or adding a separate profile for this (`Python: Django with Debug Toolbar`).

gzurowski: @yattias I updated the launch configuration to enable the debug toolbar by default.

gzurowski (approved):
LGTM 👍

PR #1817, opened 2024-08-19
Migrate paper-author relation into authorship

gzurowski:
- Add a one-off script to migrate existing `paper_paper_authors` relationships to the new `paper_authorship` table.
- Only migrated records for papers that don't have any records in `paper_authorship` yet. This is to avoid unnecessary duplicate entries (for the same author, but with a different author ID).
- All records are marked with `source=AUTHOR_MIGRATION` in the `paper_authorship`.
- Once the migration is completed, the previous `paper_paper_authors` relation can be deprecated with #1812.


yattias (approved):

src/paper/related_models/authorship_model.py
```diff
@@ -1,9 +1,13 @@
 from django.db import models
+from django.utils.translation import gettext_lazy as _
 
 from utils.models import DefaultModel
 
 
 class Authorship(DefaultModel):
+    class Source(models.TextChoices):
+        AUTHOR_MIGRATION = "AUTHOR_MIGRATION", _("Author Migration")
```
koutst, line 9: Do we assume blank entries are from OpenAlex?

koutst (approved):
LGTM

PR #1830, opened 2024-08-24
Adding job to Processing paper via OpenAlex when paper is being uploaded

yattias:


src/paper/paper_upload_tasks.py
```diff
@@ -79,9 +80,9 @@ def celery_process_paper(self, submission_id):
         apis.extend(
             [
                 celery_unpaywall.s(),
-                celery_openalex.s(),
                 celery_crossref.s(),
                 celery_semantic_scholar.s(),
+                celery_openalex.s(),
```
yattias, line 85: This ensure OpenAlex data takes priority by being added last

PR #1831, opened 2024-08-24
Add achievements details, including milestones

koutst:
- Add milestones for gold, silver and bronze achievements.

### DIFF
--- a/src/user/related_models/author_model.py
+++ b/src/user/related_models/author_model.py
@@ -233,6 +233,38 @@ class Author(models.Model):
         return achievements
 
     @property
+    def achievements_details(self):
+        upvote_count = getattr(self.user, "upvote_count", 0)
+        peer_review_count = getattr(self.user, "peer_review_count", 0)
+        amount_funded = getattr(self.user, "amount_funded", 0)
+        return {
+            "CITED_AUTHOR": {
+                "value": self.citation_count,
+                "milestones": [10, 100, 1000],
+            },
+            "OPEN_ACCESS": {
+                "value": self.open_access_pct,
+                "milestones": [0.5, 0.75, 0.875],
+            },
+            "OPEN_SCIENCE_SUPPORTER": {
+                "value": amount_funded,
+                "milestones": [1, 1000, 10000],
+            },
+            "HIGHLY_UPVOTED": {
+                "value": upvote_count,
+                "milestones": [
+                    10,
+                    100,
+                    1000,
+                ],
+            },
+            "EXPERT_PEER_REVIEWER": {
+                "value": peer_review_count,
+                "milestones": [1, 25, 250],
+            },
+        }
+
+    @property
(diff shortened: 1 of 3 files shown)


src/user/related_models/author_model.py
```diff
@@ -232,6 +232,38 @@ def achievements(self):
...
+                "value": upvote_count,
+                "milestones": [
+                    10,
+                    100,
+                    1000,
+                ],
+            },
+            "EXPERT_PEER_REVIEWER": {
+                "value": peer_review_count,
+                "milestones": [1, 25, 250],
```
yattias, line 263: Let's change the gold to 50

src/user/serializers.py
```diff
@@ -1086,6 +1087,9 @@ class Meta:
     def get_achievements(self, author):
         return author.achievements
 
+    def get_achievements_details(self, author):
```
yattias, line 1090: Given that we are trying to get out of performance hell, I think this pattern will hurt us.
Nothing wrong with the code you are writing here, but rather the pattern of how things are serialized is problematic (I think).

Explanation:
```
        upvote_count = getattr(self.user, "upvote_count", 0)
        peer_review_count = getattr(self.user, "peer_review_count", 0)
        amount_funded = getattr(self.user, "amount_funded", 0)
```

These lines are already invoked in the serializer in the `get_achievements` method. As a result, I think that we will double up our queries. 

This observation is not tested and it could be that Django is somehow smart enough to avoid doubling up the queries. Use the debug toolbar to verify. 

If it is indeed doubling up the queries, we should use another approach such as overriding the `to_representation` serializer method

koutst: I have replaced the old get_achievements instead, this will avoid the double fetching.

yattias (approved):

src/user/related_models/author_model.py
```diff
@@ -199,38 +199,32 @@ def achievements(self):
...
-        elif peer_review_count >= 1:
-            achievements.append("EXPERT_PEER_REVIEWER_1")
-
-        return achievements
+        return {
+            "CITED_AUTHOR": {
+                "value": self.citation_count,
+                "milestones": [10, 100, 1000],
+            },
+            "OPEN_ACCESS": {
```
yattias, line 207: sdfdsf

src/user/related_models/author_model.py
```diff
@@ -199,38 +199,32 @@ def achievements(self):
...
+            },
+            "OPEN_SCIENCE_SUPPORTER": {
+                "value": amount_funded,
+                "milestones": [10, 1000, 10000],
+            },
+            "HIGHLY_UPVOTED": {
+                "value": upvote_count,
+                "milestones": [
+                    10,
+                    100,
```
yattias, line 219: sdfdsfds

yattias (commented):
Test

PR #1851, opened 2024-09-16
Add description to transaction export

gzurowski:
This change adds a new _description_ column to the transaction report that shows the type of the transaction (balance) for each entry in the exported CSV file.

Additionally, write the export CSV file as a native CSV file instead of wrapping it into a JSON object.

Closes ResearchHub/issues/issues/114


koutst (approved):
LGTM

PR #1854, opened 2024-09-16
Add tests to pull_new_openalex_works

koutst:
- Add tests to pull_new_openalex_works.
- Since we're tracking next_cursor now, we no longer need logic around where to start processing.


src/researchhub/settings.py
```diff
@@ -654,6 +654,10 @@ def before_send(event, hint):
 CELERY_TASK_DEFAULT_QUEUE = "default"
 CELERY_TASK_IGNORE_RESULT = True
 
+if TESTING:
+    CELERY_TASK_ALWAYS_EAGER = True
+    CELERY_TASK_EAGER_PROPAGATES = True
+
```
gzurowski, line 660: 🎉

gzurowski (approved):
LGTM 🔥

PR #1883, opened 2024-09-27
Tiered Editors

le-sun:
researchhub/issues#115

Added simple tiering system for Hub editors.

Frontend: https://github.com/ResearchHub/researchhub-web/pull/1852


src/invite/migrations/0008_alter_noteinvitation_invite_type_and_more.py
```diff
@@ -0,0 +1,51 @@
+# Generated by Django 4.2.15 on 2024-09-26 23:33
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
yattias, line 6: What are the implications of adding these migrations? s this related to the notebook feature?
How does this change relate to new editor types?

src/researchhub_access_group/constants.py
```diff
@@ -1,5 +1,8 @@
 ADMIN = "ADMIN"
 EDITOR = "EDITOR"
+ASSISTANT_EDITOR = "ASSISTANT_EDITOR"
```
yattias, line 3: Are we able to display the editor status on profile page? Can you recommend an approach for us to do it? 

e.g Leo sun is Associate Editor of Neurology

yattias (changes requested):
@le-sun Thanks for this PR!
Since this is very critical for us, can you add some test to ensure that assistant, associate and senior editors get paid the amount expected? Thank you

PR #1885, opened 2024-09-30
Put paper creation in transaction to avoid creation of paper without unified document

koutst:
- Unified documents are created in post_save signals. If they fail we want the paper creation to fail. Putting paper creation in a transaction should do that.


gzurowski (approved):
LGTM 👍

PR #1894, opened 2024-10-02
Unified Document view optimization

yattias:
- Remove unnecessary hub related filtering code that appears to cause massive slowness
- Adding `is_used_for_rep` flag


gzurowski (approved):
LGTM 👍

PR #1918, opened 2024-10-16
Debounce signals for registry update tasks

gzurowski:
Prevent multiple Elasticsearch reindex requests on the same model instance within the specified `DEBOUNCE_SIGNAL_PERIOD`.

The following screenshot shows updates triggered for one paper within the same second:
image

### DIFF
--- a/src/search/celery.py
+++ b/src/search/celery.py
@@ -1,27 +1,38 @@
 from celery import shared_task
 from django.apps import apps
+from django.core.cache import cache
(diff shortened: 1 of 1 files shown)


yattias (approved):

gzurowski:
Verified in staging that only one signal gets processed for multiple `save` operations on the same papers (without the change, updating the title of a paper triggered three signals).

image

yattias (approved):

PR #1919, opened 2024-10-16
Filtering for bounties page; Additional metadata

yattias:
- Adding filtering logic to bounties list endpoint: `hub_ids`, and `bounty_type` params are supported
- Adding additional metadata such as `hubs` and `abstract`
- Adding official account marker to indicate official accounts from community ones

Related to: https://github.com/ResearchHub/researchhub-web/pull/1861


src/reputation/views/bounty_view.py
```diff
@@ -426,22 +433,60 @@ def cancel_bounty(self, request, pk=None):
...
+            applied_filters &= Q(created_by__in=researchhub_official_user_accounts)
+
+        # Handle review, answer, and other filters
+        review_or_answer_filter = Q()
+        if "review" in bounty_types:
+            review_or_answer_filter |= Q(bounty_type="REVIEW")
+        if "answer" in bounty_types:
+            review_or_answer_filter |= Q(bounty_type="ANSWER")
+        if "other" in bounty_types:
+            review_or_answer_filter |= Q(bounty_type="OTHER")
```
gzurowski, line 462: Does a `OTHER` bounty type exist?

src/user/related_models/user_model.py
```diff
@@ -96,6 +96,10 @@ class User(AbstractUser):
     updated_date = models.DateTimeField(auto_now=True)
     upload_tutorial_complete = models.BooleanField(default=False)
 
+    # Official accounts are currently used as a distinguishing factor in bounties.
+    # Bounties are separated into two categories: official and community.
+    is_official_account = models.BooleanField(default=False)
```
gzurowski, line 101: Nit: Can we find a better name than "official"?

yattias: Official is the best name I could come up with. Do you have a suggestion?

src/reputation/views/bounty_view.py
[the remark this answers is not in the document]
yattias: Done

yattias: It does, but it is called "GENERIC_COMMENT" in the database 🫤

src/reputation/related_models/bounty.py
```diff
@@ -36,6 +36,14 @@ class Bounty(DefaultModel):
...
         (EXPIRED, EXPIRED),
     )
+    REVIEW_TYPE = "REVIEW"
+    ANSWER_TYPE = "ANSWER"
+    OTHER_TYPE = "GENERIC_COMMENT"
+    type_choices = (
+        (REVIEW_TYPE, REVIEW_TYPE),
+        (ANSWER_TYPE, ANSWER_TYPE),
+        (OTHER_TYPE, OTHER_TYPE),
+    )
```
gzurowski, line 46: Nit: Should we use `TextChoice` here?

Example: https://github.com/ResearchHub/researchhub-backend/blob/fde9c45981bc966d6f6278ce21de4a7f6022e162/src/user/related_models/user_verification_model.py#L14

src/reputation/views/bounty_view.py
```diff
@@ -439,27 +439,29 @@ def filter_queryset(self, queryset):
...
         applied_filters = Q()
 
         # Only return bounties within specific hubs
         if hub_ids:
+            print("hub_ids", hub_ids)
             applied_filters &= Q(unified_document__hubs__id__in=hub_ids)
 
         # ResearchHub foundation only filter
-        if "researchhub" in bounty_types:
+        if "RESEARCHHUB" in bounty_types:
```
gzurowski, line 451: Should we also add `RESEARCHHUB` as a bounty type choice?

yattias: "RESEARCHHUB" as a filter type makes sense in the context of the bounties page to the user

From an API perspective, it should not be stored as a type in the database because it refers to a bounty that is created by the researchhub community. It is not a "type" per se, but from the user perspective, it makes sense:

image

src/reputation/related_models/bounty.py
yattias: Made change

PR #1932, opened 2024-10-22
feat: Create new API for uploading papers

gzurowski:
Create a new API (`/paper/upload`) for creating pre-signed URLs, so that the caller can upload papers directly to S3. This will be the first step when submitting new papers to RH.

Closes https://github.com/ResearchHub/issues/issues/131


yattias (approved):
Looks good. How would the client use this presigned url?

PR #1933, opened 2024-10-22
Add versions paper model serializer

koutst:
- Closes https://github.com/ResearchHub/issues/issues/136


src/paper/serializers/paper_serializers.py
```diff
@@ -1037,6 +1062,30 @@ def get_pdf_url(self, paper):
             return paper.pdf_url
         return None
 
+    def get_version(self, paper):
+        print("GETTING VERSION DYNAMIC")
```
yattias, line 1066: I imagine we can remove this

yattias (approved):

src/paper/serializers/paper_serializers.py
```diff
@@ -1037,6 +1062,29 @@ def get_pdf_url(self, paper):
...
 
+    def get_version(self, paper):
+        try:
+            paper_version = PaperVersion.objects.get(paper=paper)
+        except PaperVersion.DoesNotExist:
+            return None
+
+        return paper_version
+
+    def get_version_list(self, paper) -> list:
```
gzurowski, line 1073: Nit:

```suggestion
    def get_versions(self, paper) -> list:
```

src/paper/serializers/paper_serializers.py
```diff
@@ -1037,6 +1062,29 @@ def get_pdf_url(self, paper):
...
+        try:
+            paper_version = PaperVersion.objects.get(paper=paper)
+        except PaperVersion.DoesNotExist:
+            return None
```
gzurowski, line 1077: Returning an empty list makes it more convenient for the caller (and also evaluates falsely).

```suggestion
            return []
```

gzurowski (approved):

src/paper/serializers/paper_serializers.py
koutst: The field is called version_list to stay consistent with other list fields.

PR #1961, opened 2024-11-04
Add DOI class

koutst:
Closes: https://github.com/ResearchHub/issues/issues/145

### DIFF
--- a/src/utils/doi.py
+++ b/src/utils/doi.py
@@ -6,14 +6,7 @@ from datetime import datetime
 import requests
 from django.template.loader import render_to_string
 
-from researchhub.settings import (
-    BASE_FRONTEND_URL,
-    CROSSREF_API_URL,
-    CROSSREF_DOI_PREFIX,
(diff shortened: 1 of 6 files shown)


src/utils/doi.py
```diff
@@ -0,0 +1,86 @@
...
+from django.template.loader import render_to_string
+
+from researchhub.settings import (
+    BASE_FRONTEND_URL,
+    CROSSREF_API_URL,
+    CROSSREF_DOI_PREFIX,
+    CROSSREF_DOI_SUFFIX_LENGTH,
+    CROSSREF_LOGIN_ID,
+    CROSSREF_LOGIN_PASSWORD,
+)
```
gzurowski, line 16: Don't import setting directly, import the `settings` object instead.

src/utils/doi.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class DOI:
```
gzurowski, line 19: Maybe add some tests?

gzurowski (approved):

PR #1989, opened 2024-11-12
Fix tip division on paper

koutst:
Closes: https://github.com/ResearchHub/issues/issues/143 
- The issue with reward is a formatting confusion caused by the use of a comma as a decimal separator which arises because of the use of toLocaleString.

### DIFF
--- a/src/purchase/utils.py
+++ b/src/purchase/utils.py
@@ -8,7 +8,6 @@ from reputation.models import Escrow
 def distribute_support_to_authors(paper, purchase, amount):
     registered_authors = paper.authorship_authors.all()
     total_author_count = registered_authors.count()
-    print("total_author_count", total_author_count)
 
     rewarded_rsc = 0
     if total_author_count:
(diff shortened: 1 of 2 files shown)


src/purchase/utils.py
```diff
@@ -7,7 +7,8 @@
 
 def distribute_support_to_authors(paper, purchase, amount):
     registered_authors = paper.authorship_authors.all()
-    total_author_count = paper.true_author_count()
+    total_author_count = registered_authors.count()
+    print("total_author_count", total_author_count)
```
yattias, line 11: remove?

gzurowski (approved):
LGTM 🚀

Carlostt-ux:
should i reapply for the paper award?

which I tried and it does not allow it

the RSC amount has not changed

should I wait?

Carlostt-ux:
The problem with my account is still persistent, the RSCs corresponding to the award for my paper and the division of the tip have not been credited.

PR #2017, opened 2024-11-26
chore: Use AWS utilities instead of smart open

gzurowski:
Stop using smart open for accessing S3 objects. Instead, use the AWS client from the utility package.


yattias (approved):
Nice! we should test notebook works since it uses AWS for storage

PR #2029, opened 2024-12-03
chore: Remove unused `postgres` package

gzurowski:


src/paper/tasks.py
```diff
@@ -223,7 +222,7 @@ def celery_extract_meta_data(paper_id, title, check_title):
 
         paper.check_doi()
         paper.save()
-    except (UniqueViolation, IntegrityError) as e:
+    except IntegrityError as e:
```
koutst, line 225: Can we remove this except since it does the same thing as the except below it?

koutst (approved):

PR #2033, opened 2024-12-06
chore: Add index to `openalex_ids` array fields

gzurowski:


src/user/related_models/author_model.py
```diff
@@ -104,6 +105,11 @@ class Author(models.Model):
         related_name="merged_authors",
     )
 
+    class Meta:
+        indexes = [
+            GinIndex(fields=["openalex_ids"], name="openalex_ids_idx"),
```
yattias, line 110: Never heard of GinIndex. Read about it just now - pretty cool

PR #2041, opened 2024-12-11
ACK after long running tasks complete

gzurowski:
The default behavior is to ACK tasks when they start. Sending an ACK at the end of processing allows the task to be picked up again in scenarios where the celery broker crashes.

### DIFF
--- a/src/paper/openalex_tasks.py
+++ b/src/paper/openalex_tasks.py
@@ -18,7 +18,7 @@ from utils.openalex import OpenAlex
 logger = logging.getLogger(__name__)
 
 
-@app.task(bind=True, max_retries=3)
+@app.task(bind=True, late_acks=True, max_retries=3)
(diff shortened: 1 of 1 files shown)


koutst (approved):
LGTM. Could this lead to the task running multiple times at once?

yattias:
@gzurowski This is really interesting. So adding `late_acks` would not end the task?

Also, @koutst comment is pretty important

yattias (approved):

gzurowski:
> LGTM. Could this lead to the task running multiple times at once?

This is prevented by the locking mechanism added with #2043.

gzurowski:
> @gzurowski This is really interesting. So adding `late_acks` would not end the task?

This is so a previously running task that was interrupted by deployments, crashes, etc. gets picked up again.

PR #2046, opened 2024-12-14
fix: minor setup config updates for easy installation

fatherabraham-hms:
Added missing keys and updated db load script to remove incorrect spacing.


gzurowski (approved):
LGTM 👍 

Thanks for your contribution.

PR #2049, opened 2024-12-17
Support base withdrawal

koutst:
- Add support for Base withdrawals, keep Ethereum support.


src/researchhub/settings.py
```diff
@@ -673,6 +673,22 @@ def before_send(event, hint):
...
+WEB3_BASE_PROVIDER_URL = os.environ.get(
+    "WEB3_BASE_PROVIDER_URL", keys.WEB3_BASE_PROVIDER_URL
+)
+
+# Create Web3 instances for each network
+w3_ethereum = Web3(Web3.HTTPProvider(WEB3_PROVIDER_URL))
+w3_base = Web3(Web3.HTTPProvider(WEB3_BASE_PROVIDER_URL))
+
+# For backward compatibility
+w3 = w3_ethereum
```
gzurowski, line 691: Nit: Probably outside the scope of this change - I am wondering if there is a way to remove these global instances from the settings file? The settings file should ideally hold settings derived from configuration, not host class instances that should be colocated with the actual consumers of such functionality.

gzurowski (approved):
LGTM 👍

koutst:
@gzurowski I will address your comments in a follow up deposit PR

PR #2055, opened 2024-12-19
Get gas price from basescan

koutst:


src/reputation/views/withdrawal_view.py
```diff
@@ -28,15 +28,19 @@
 from reputation.models import Withdrawal
 from reputation.permissions import AllowWithdrawalIfNotSuspecious
 from reputation.serializers import WithdrawalSerializer
-from researchhub.settings import ETHERSCAN_API_KEY, WEB3_PROVIDER_URL, WEB3_RSC_ADDRESS
+from researchhub.settings import (
+    BASESCAN_API_KEY,
+    ETHERSCAN_API_KEY,
+    WEB3_PROVIDER_URL,
+    WEB3_RSC_ADDRESS,
+)
```
gzurowski, line 36: Import `settings` not the settings themselves.

src/reputation/views/withdrawal_view.py
```diff
@@ -214,11 +226,12 @@ def calculate_transaction_fee(self, network="ETHEREUM"):
             )
             json = res.json()
             gas_price = json.get("result", {}).get("SafeGasPrice", 40)
-            gas_limit = 120000.0
+
+        gas_limit = 120000.0
```
gzurowski, line 230: Nit: Does it make sense to add a comment or use a constant for this magic number?

gzurowski (approved):
LGTM 👍

src/reputation/views/withdrawal_view.py
koutst: I have an upcoming PR that is based on this and fixes it.

PR #2056, opened 2024-12-20
Changing pfp length to avoid authentication error (& ORCID text)

TylerDiorio:
Changed profile picture length of field to 2048 to avoid long Google URL links for profile pictures (ex. URL was 1042 characters in length).
- Not possible to easily change type from FileType to text unfortunately.

Also changed ORCID to text-based field instead of character.


koutst (approved):
LGTM. Is there a reason we removed the uniqueness constraint for the orcid_id?

koutst:
Sorry, you also need to run makemigrations to create the actual migration that accompanies these changes.

PR #2067, opened 2025-01-10
TurboTax csv export + filters on exchange_rate

TylerDiorio:
- New endpoint in balance_view "list_csv_turbotax" which gives a pull of all transactions from the user with exchange rates at the time of tx, formatted specifically for turbotax
  - Improved transaction type mapping for tax reporting
- Added filters on exchange rate (created_date and price_source)
- Maintained backward compatibility with original CSV export


src/purchase/views/balance_view.py
```diff
@@ -117,3 +165,86 @@ def list_csv(self, request):
             )
 
         return response
+
+    @action(
+        detail=False,
+        methods=["GET"],
+        permission_classes=[IsAuthenticated],
+    )
+    def list_csv_turbotax(self, request):
```
yattias, line 174: Let's call this method `turbotax_csv_export` to be more specific

src/user/related_models/author_model.py
yattias, line 1: These changes are already on master

https://github.com/ResearchHub/researchhub-backend/pull/2056/files

src/notification/views.py
```diff
@@ -71,13 +72,22 @@ def partial_update(self, request, *args, **kwargs):
...
+        Notification.objects.filter(recipient=user).update(
             read=True, read_date=timezone.now()
         )
         return Response("Success", status=status.HTTP_200_OK)
 
+    @action(
+        detail=False,
+        methods=["GET"],
+    )
+    def unread_count(self, request):
```
yattias, line 85: This one is my code from my notification branch. OK to merge though!

src/notification/tests.py
```diff
@@ -0,0 +1,54 @@
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APIClient, APITestCase
+
+from notification.models import Notification
+from paper.tests.helpers import create_paper
+from user.tests.helpers import create_random_default_user
+
+
+class NotificationViewSetTests(APITestCase):
```
yattias, line 10: This one is my code from my notification branch. OK to merge though!

gzurowski (changes requested):
Please remove any unrelated changes (notification, unrelated database migration).

src/purchase/views/balance_view.py
```diff
@@ -117,3 +165,86 @@ def list_csv(self, request):
...
+
+            transaction_type = self.get_transaction_type(balance)
+            amount = abs(Decimal(balance.amount))  # Use absolute value for amounts
+            usd_value = amount * Decimal(rate)
+            
+            # Determine if this is a send or receive based on amount sign
+            is_negative = Decimal(balance.amount) < 0
+            
+            # Format the row based on transaction type and amount sign
+            if is_negative or transaction_type in ["Withdrawal", "Expense"]:
```
koutst, line 215: Won't withdrawals and expenses (fees) always be negative? Do they need to be treated differently than for example purchases?

src/purchase/views/balance_view.py
```diff
@@ -67,6 +80,41 @@ def send_rsc(self, request):
 
         return Response({"message": "RSC Sent!"})
 
+    @staticmethod
+    def get_transaction_type(balance) -> str:
+        """
+        Map balance content type to TurboTax transaction type.
```
koutst, line 86: Could we add the Turbotax docs? Could make it easier when updating this feature later to have them here.

src/purchase/views/balance_view.py
TylerDiorio: I saw some older withdrawals that have positive values (436 positive value withdraws, most recently as 02/26/2022). Looks like all supportfees and bountyfees are negative though so I can remove the "Expense" there - thanks Taki

> Won't withdrawals and expenses (fees) always be negative? Do they need to be treated differently than for example purchases?

I saw some older withdrawals that have positive values (436 positive value withdraws, most recently as 02/26/2022). 
Looks like all supportfees and bountyfees are negative though so I can remove the "Expense" there - thanks Taki

gzurowski (approved):
LGTM 👍

kmk142789 (approved):

PR #2071, opened 2025-01-15
Format ORCID ID to match XXXX-XXXX-XXXX-XXXX pattern and strip URL if present

koutst:


gzurowski (approved):
LGTM

Since the `register_doi_for_paper` method is already pretty long, it might be worth extracting the new ORCID code into a separate function (e.g. `sanitizeOrcid`) to avoid adding more conditional logic.

PR #2073, opened 2025-01-16
chore: Remove old follow feature

gzurowski:
This change removes the old "follow" feature that was limited to following users. The feature has been never used (there is a single row in the corresponding production database).

**NOTE**

- [x] #2072 must be merged first.


yattias (approved):

koutst (approved):
LGTM

PR #2076, opened 2025-01-17
chore: Remove latest activities

gzurowski:
Remove the `following_latest_activity` endpoint and the related signal and task logic. `following_latest_activity` is not used by the frontend and the existing logic populates/invalidates a cache that is not consumed.

FE changes are removed with https://github.com/ResearchHub/researchhub-web/pull/1904.


koutst (approved):
LGTM

yattias (approved):

PR #2081, opened 2025-01-20
Combined Suggesting papers from OpenAlex and ElasticSearch 

yattias:
- Suggesting papers from OpenAlex and ElasticSearch via new suggest endpoint
- Endpoint `retrieve_by_doi` which will retrieve paper if exists or fetch it from OpenAlex and upload it to our DB


src/paper/views/paper_views.py
```diff
@@ -1086,6 +1097,90 @@ def edit_file_extract(self, request, pk=None):
         )
         return Response(status=status.HTTP_200_OK)
 
+    @action(
+        detail=False,
+        methods=["get"],
+        permission_classes=[AllowAny],
+    )
+    def retrieve_by_doi(self, request):
```
yattias, line 1105: This will synchronously process paper from openalex if not in our database. 
LMK if anyone has better ideas of how to do this

gzurowski (approved):
Looks good except for lack of tests.

PR #2088, opened 2025-01-28
Create Fundraise on Preregistration Creation 

nicktytarenko:
This PR adds the ability to create fundraising campaigns during preregistration post creation. 

## Changes
- Added fundraise creation functionality to the existing post creation endpoint
- Created helper function `create_fundraise_with_escrow` to handle fundraise and escrow creation so we can reuse it in the post_viewes
- Notes:
  - Fundraise creation only triggers if `fundraise_goal_amount` property exists in the request


nicktytarenko:
In the web app we will be able to send the following payload to their endpoint

```typescript
interface CreatePostPayload {
  // Required fields with validation
  title: string; 
  renderable_text: string;
  document_type: DocumentType;
  full_src: string;

  // Optional fields
  authors?: Author[];
  note_id?: number | null;
  editor_type?: EditorType;
  assign_doi?: boolean;
  preview_img?: string;
  bounty_type?: string;
  hubs?: number[];

  fundraise: {
    goal_amount: number;
    goal_currency: 'USD';
  };
}
```

Also, there is validation, and would be great to mirror this validation on the FE app, so we do not send the invalid data, for example:
```
  if (!payload.title || payload.title.length < 20) {
    errors.push('Title must be at least 20 characters');
  }

  if (!payload.renderable_text || payload.renderable_text.length < 50) {
    errors.push('Post body must be at least 50 characters');
  }
  if (payload.document_type !== 'PREREGISTRATION') {
      errors.push('Fundraise can only be created for PREREGISTRATION type');
    }
    if (payload.fundraise.goal_amount <= 0) {
      errors.push('Fundraise goal amount must be greater than 0');
    }
    if (payload.fundraise.goal_currency !== 'USD') {
      errors.push('Only USD currency is supported for fundraise');
    }
```

src/purchase/utils.py
```diff
@@ -31,3 +39,59 @@ def store_leftover_paper_support(paper, purchase, leftover_amount):
         item=paper,
         hold_type=Escrow.AUTHOR_RSC,
     )
+
+
+def create_fundraise_with_escrow(
```
gzurowski, line 44: This is not a comment requiring immediate action but I wanted to share a suggestion: For new apps, we should avoid putting business logic into utility packages (`utils.py`). For better cohesion (and testability), it would be preferable to organize business logic in service packages or classes as done in https://github.com/ResearchHub/researchhub-backend/blob/555c0feedd6ca52382f3d8bf605c4f9957d9c505/src/paper/services/storage_service.py#L14 Since `purchase` is an older app, we can continue as it.

src/purchase/utils.py
```diff
@@ -31,3 +39,59 @@ def store_leftover_paper_support(paper, purchase, leftover_amount):
...
+    user, unified_document, goal_amount, goal_currency=USD, status=Fundraise.OPEN
```
gzurowski, line 45: It would be helpful to add type hints to the function signature.

src/purchase/utils.py
```diff
@@ -31,3 +39,59 @@ def store_leftover_paper_support(paper, purchase, leftover_amount):
...
+):
+    """
+    Helper function to create a fundraise with its associated escrow.
+    Returns (fundraise, error_response) tuple where error_response is None if successful
+
+    Note: This function should be called within a transaction.atomic() block
+    """
+    # Validate inputs
+    if unified_document.document_type != PREREGISTRATION:
+        return None, Response(
```
gzurowski, line 55: This code mixes business logic with presentation logic. It would be cleaner to have the function return values or raise exceptions that are independent of the HTTP view layer. The fact that `rest_framework` is imported here is an indicator for inappropriately mixing these two aspects.

gzurowski (approved):
LGTM 👍 

Thanks for implementing the suggested changes.

I made another comment regarding injecting the service into the view instead of hard-coding it within the `create` method of the view.

Since we have a good amount of validation logic, we might also want to fail fast and perform validation in the view (before even starting a database transaction, etc.). In Django such validation logic can be managed in a serializer.

You could create a `class FundraiseCreateSerializer(serializers.ModelSerializer)` and put all validation logic within the `def validate(...)` method (there is also the `serializers.ValidationError` for signaling validation errors). You can then hook in the serializer as such:
```
def create(self, request, *args, **kwargs):
    serializer = FundraiseCreateSerializer(data=request.data)
    serializer.is_valid(raise_exception=True)
```

nicktytarenko:
hey @gzurowski it was not easy as I thought) but it looks like now all tests passed and everything works as expected.
Could you please take a look again
thank you!

src/purchase/tests/test_fundraise_service.py
```diff
@@ -0,0 +1,176 @@
...
+            mock_create_fundraise.assert_called_once_with(
+                created_by=self.user,
+                unified_document=self.unified_document,
+                goal_amount=goal_amount,
+                goal_currency=USD,
+                status=Fundraise.OPEN,
+            )
+
+            mock_create_escrow.assert_called_once()
+            self.assertEqual(result, mock_fundraise)
```
gzurowski, line 63: The excessive amount of patching indicates that the test is not ideal. It would be more readable and maintainable if the objects get created and verified.

gzurowski (approved):
LGTM

Added two comments about an unused custom exception and an overuse of patching in a test.

src/purchase/services/fundraise_service.py
```diff
@@ -0,0 +1,54 @@
...
+
+from purchase.models import Fundraise
+from purchase.related_models.constants.currency import USD
+from reputation.models import Escrow
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+
+
+class FundraiseValidationError(Exception):
```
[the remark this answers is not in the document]
nicktytarenko: ohh, you are right. we raise `ValidationError` in the serializer
will remove it

src/purchase/tests/test_fundraise_service.py
nicktytarenko: agree 100%

nicktytarenko:
> LGTM
> 
> Added two comments about an unused custom exception and an overuse of patching in a test.

Thanks @gzurowski could you please review the latest commit of the PR?

gzurowski:
> > LGTM
> > Added two comments about an unused custom exception and an overuse of patching in a test.
> 
> Thanks @gzurowski could you please review the latest commit of the PR?

Thanks for addressing all comments. LGTM. 👍

PR #2091, opened 2025-02-03
HTML column on comments for storing tiptap editor data

yattias:
- To support tiptap, we need to introduce a new html property
- For backwards compatibility, we will still support comment_content_json


gzurowski (approved):
LGTM 👍

PR #2094, opened 2025-02-04
Nest user in author, add action_date, distinct, only add papers with published date

koutst:
- Recreate FeedEntry table, adding action_date (corresponding to paper published date)
- Nest user in author serializer
- Distinct on unique items
- Do not add to Feed if no publish date


src/paper/signals.py
```diff
@@ -61,6 +61,9 @@ def handle_paper_hubs_changed(sender, instance, action, pk_set, **kwargs):
                 hub = instance
                 paper = hub.papers.get(id=hub_id)
 
+            if paper.paper_publish_date is None:
```
gzurowski, line 64: I would be useful to have an additional comment here. E.g.:
> Papers without a publish date are not displayed in the feed.

gzurowski (approved):
LGTM 👍

PR #2102, opened 2025-02-10
Add feed signal handlers for bounties

gzurowski:
Add a new bounty signal handler for creating feed entries when bounties are opened.


yattias (approved):

koutst (approved):
LGTM

PR #2110, opened 2025-02-13
fix: Reuse and add prefetched querysets for feed

gzurowski:
- Reuse prefetched querysets (instead of creating new querysets when calling `.first()`.
- Add prefetch for author's user relation.

=> Reduces the number of queries for a single feed page from 78 to 6 queries.


koutst (approved):

yattias (approved):
Nice optimization

PR #2145, opened 2025-02-28
feat: Add unified document to feed entries

gzurowski:
Add a direct relation to unified documents to feed entries.
This is primarily for performance reasons. Since different entity types managed in the feed can have their relation to the unified document on a different level, this change makes sure that we can easily join the feed with the unified document uniformly.


src/feed/models.py
```diff
@@ -34,6 +37,15 @@ class FeedEntry(DefaultModel):
     action_date = models.DateTimeField(db_index=True)
     user = models.ForeignKey(User, on_delete=models.CASCADE, null=True, blank=True)
 
+    unified_document = models.ForeignKey(
+        ResearchhubUnifiedDocument,
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
```
koutst, line 44: Should we make it a non nullable field or do we want to avoid creating that sort of requirement on feed entries?

koutst (approved):

PR #2146, opened 2025-02-28
Remove v1 of hot_score from unified document

koutst:
- The hot_score that is actually used is hot_score_v2
- Rename hot_score_v2 to hot_score


gzurowski (approved):
LGTM 🧹

PR #2171, opened 2025-03-10
Feat/endaoment integration

TylerDiorio:
• Created API endpoint at /api/nonprofit_orgs/search/ to proxy Endaoment API
• Flexible parameter handling with support for all Endaoment search options
• Added sensible defaults (count=15, offset=0) for pagination parameters
• Test suite with structure-based validation
• Added detailed documentation for maintainability


src/nonprofit/__init__.py
gzurowski, line 1: On the root `src` level we manage Django apps with the exception of `utils`. The `utils` package is a bit generic, but I rather see the `nonprofit` package there than directly under `src`.

gzurowski, line 1: I am wondering where to locate this best since we usually manage Django apps on the `src` root level (with the exception of the `utils` package). `nonprofit` seems like a tiny proxy entity within our system and it doesn't seem to register itself as a Django app. Should we have a subfolder for these types of integrations?

src/utils/endaoment.py
```diff
@@ -0,0 +1,64 @@
+from typing import Any, Dict, Optional
+from urllib.parse import urlencode
+
+import requests
+
+
+def search_nonprofit_orgs(
```
gzurowski, line 7: This is not located appropriately. The view is in `nonprofit` but the core functionality resides in the generic `utils` package. Ideally this is co-located with the rest of the code.

src/utils/endaoment.py
```diff
@@ -0,0 +1,64 @@
...
+    except requests.RequestException as e:
+        # Log error and return error response
+        return {
+            "error": str(e),
+            "status": (
+                getattr(e.response, "status_code", 500)
+                if hasattr(e, "response")
+                else 500
+            ),
+        }
```
gzurowski, line 64: The comment states that some logging is performed, but it's not. We can easily log at an appropriate level with the `logging` package.

Additional nit-pick: This comment is not very useful since it only describes what the simple code below already expresses. Additionally, such comments get easily out of sync when the code evolves.

gzurowski (commented):
Generally fine, but I suggest to find a better location for the new code, also to improve cohesion by co-locating all parts.

src/nonprofit_orgs/__init__.py
gzurowski, line 1: Should we name this app `organizations` where non-profit is one entity within the app?

gzurowski (approved):
LGTM

PR #2198, opened 2025-03-20
Extract feed metrics into separate field

gzurowski:
This is mainly a performance improvement change to avoid querying metrics when the feed is queried. Metrics are stored with the feed records and updated with signals instead.

- Extract feed metrics into its own `metrics` field (metrics are more dynamic than the actual item content).
- Add new JSON field `metrics` to the feed entries model.
- Populate metrics (`votes`) via triggered signals and tasks when votes are submitted towards papers, posts, or comments.
- Populate metrics (`replies`) via comments signals and tasks when comments are made.


yattias (approved):

koutst (approved):

koutst:
Should we add a command for recreating the metrics and content fields?

gzurowski:
> Should we add a command for recreating the metrics and content fields?

https://github.com/ResearchHub/researchhub-backend/pull/2198/commits/94f9d07c3a47dd92ba4e7fb107c7c7e2991cb798

koutst (approved):

PR #2200, opened 2025-03-25
Improve popular feed tab, don't distinct on unified doc id

koutst:
- We are currently apply distinct on unified document id of feed entries. This will cause the most recent entry for that unified doc to show.
- We should distinct the same way we do the latest tab, the actual id of the entity.


yattias (approved):

gzurowski (approved):
LGTM 👍

PR #2203, opened 2025-03-26
Add funding feed

koutst:
- Creates a pseudo feed built by getting preregistration posts and instantiating feed entries for those posts.
- We took this approach for performance benefits while still providing the front end with a similar structure to the general feed.


src/feed/views/funding_feed_view.py
```diff
@@ -0,0 +1,187 @@
...
+                    )
+                elif fundraise_status.upper() == "CLOSED":
+                    queryset = queryset.filter(
+                        Q(unified_document__fundraises__status=Fundraise.CLOSED)
+                        | Q(unified_document__fundraises__status=Fundraise.COMPLETED)
+                    )
+            except Exception as e:
+                # Log the exception but continue - useful for tests
+                print(f"Error filtering by fundraise_status: {e}")
+                pass
```
gzurowski, line 183: Use logging instead

koutst: Sorry, this was for debugging. Will remove.

gzurowski (approved):

PR #2229, opened 2025-04-02
Add materialized view for feeds

gzurowski:
- Add materialized views for the feed (popular and latest) implemented as non-managed models in Django.
- Update the existing Django feed view to use the materialized view for querying.


koutst (approved):
LGTM!

yattias (approved):

PR #2253, opened 2025-04-15
Remove join on unified document from popular feed view

gzurowski:
The feed entries table has its own `hot_score` column (both for performance reasons and for providing hot scores to feed entries other than papers or posts). This change removes the join on the unified document table that was only needed to fetch the hot score for each feed entry.


yattias (approved):

koutst (approved):
LGTM, seems like some tests need updating

PR #2288, opened 2025-04-26
Fix reviews in feed

koutst:
- Need another signal for updating feed review comments.
- The comment is created and then a review on the unified document so on the comment creation the review doesn't exist yet.

### DIFF
--- a/src/feed/tests/signals/test_review_signals.py
+++ b/src/feed/tests/signals/test_review_signals.py
@@ -1,0 +1,267 @@
+import logging
+from unittest.mock import MagicMock, patch
+
+from django.contrib.contenttypes.models import ContentType
+from django.test import TestCase, override_settings
+
+from feed.models import FeedEntry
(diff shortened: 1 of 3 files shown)


src/feed/tests/signals/test_review_signals.py
```diff
@@ -0,0 +1,262 @@
...
+)
+from researchhub_document.models import ResearchhubPost
+from researchhub_document.related_models.constants import document_type
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+from review.models.review_model import Review
+from user.tests.helpers import create_random_default_user
+
+logger = logging.getLogger(__name__)
```
gzurowski, line 26: Nit: Is the logger used in this test?

gzurowski (approved):
LGTM 👍

yattias (approved):

PR #2293, opened 2025-05-01
Endpoint for getting hubs for open bounties

yattias:
- Get hubs associated with open bounties
- These hubs will be used as filters in the Earn page
- Filtering by hub already works but issue is that we are currently returning all hubs and it is a guessing game from the user to fetch the ones that are tagged with bounty


koutst (approved):
LGTM but is the caching for too long? Can't it lead to missing hubs to filter on for 24 hours after bounty creation?

yattias:
@koutst yeah was thinking of lowering it to one hour

adamandrews1 (approved):

PR #2294, opened 2025-05-01
Add Auth Token to Email Verification Endpoint Response

nicktytarenko:
This PR enhances the `/api/auth/register/verify-email/` endpoint to return an authentication token (key) in the response upon successful email verification. This allows clients to automatically log in users immediately after they verify their email, streamlining the onboarding process.

image

### DIFF
--- a/src/user/views/custom_verify_email_view.py
+++ b/src/user/views/custom_verify_email_view.py
@@ -1,30 +1,60 @@
 from allauth.account.models import EmailConfirmation, EmailConfirmationHMAC
 from dj_rest_auth.registration.views import VerifyEmailView
+from django.contrib.auth import get_user_model
 from rest_framework.authtoken.models import Token
 from rest_framework.response import Response
+
+User = get_user_model()
 
 
 class CustomVerifyEmailView(VerifyEmailView):
     def post(self, request, *args, **kwargs):
+
         key = request.data.get("key")
+
+        if not key:
+            return Response({"detail": "Invalid key"}, status=400)
+
+        # Store user info BEFORE calling parent method
+        # because the parent method will override it
+        user_id = None
+
         confirmation = EmailConfirmationHMAC.from_key(key)
+
         if not confirmation:
-            # fallback to DB lookup (for older keys)
             confirmation = EmailConfirmation.objects.filter(key=key.lower()).first()
+
         if confirmation:
-            confirmation.confirm(request)
             user = confirmation.email_address.user
-            token, _ = Token.objects.get_or_create(user=user)
-            return Response(
-                {
-                    "detail": "ok",
-                    "key": token.key,
-                    "user": {
-                        "id": user.id,
-                        "email": user.email,
-                        "first_name": user.first_name,
-                        "last_name": user.last_name,
-                    },
-                }
-            )
-        return Response({"detail": "Invalid key"}, status=400)
+            user_id = user.id
+
+        # Call parent method to handle verification
+        parent_response = super().post(request, *args, **kwargs)
+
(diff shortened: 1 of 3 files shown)


src/user/views/custom_verify_email_view.py
```diff
@@ -0,0 +1,60 @@
...
+                        "key": token.key,
+                        "user": {
+                            "id": user_id,
+                        },
+                    }
+                )
+            except Exception as e:
+                # Log the error but don't raise it - just continue to return parent response
+                print(f"NOTE: Could not generate token response: {str(e)}")
+                print("Falling back to parent response.")
```
nicktytarenko, line 54: do we need to log it somewhere?

src/user/views/custom_verify_email_view.py
```diff
@@ -0,0 +1,60 @@
...
+                    {
+                        "detail": "ok",
+                        "key": token.key,
+                        "user": {
+                            "id": user_id,
+                        },
+                    }
+                )
+            except Exception as e:
+                # Log the error but don't raise it - just continue to return parent response
```
yattias, line 52: @koutst any suggestion of where to log it?

yattias (approved):

src/user/views/custom_verify_email_view.py
gzurowski, line 54: Remove `print` statements and replace them with regular logging.
- Import logging: `import logging`
- Create logger at the top of the file: `logger = logging.getLogger(__name__)`
- Use `logger` as usual, e.g. `logger.warning(f"...")`

src/user/views/custom_verify_email_view.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if confirmation:
+            user = confirmation.email_address.user
+            user_id = user.id
+
+        # Call parent method to handle verification
+        parent_response = super().post(request, *args, **kwargs)
+
+        # If parent verification was successful AND we have user info
+        if parent_response.status_code == 200 and user_id:
```
gzurowski, line 35: Nit: Instead of nesting the main logic by the conditional `if` and the `try`/`except`, it might be easier to return here in case of errors ("align happy path to the left").

src/user/views/custom_verify_email_view.py
```diff
@@ -0,0 +1,60 @@
+from allauth.account.models import EmailConfirmation, EmailConfirmationHMAC
+from dj_rest_auth.registration.views import VerifyEmailView
+from django.contrib.auth import get_user_model
+from rest_framework.authtoken.models import Token
+from rest_framework.response import Response
+
+User = get_user_model()
+
+
+class CustomVerifyEmailView(VerifyEmailView):
```
gzurowski, line 10: Nit: Add some comments about the purpose of this view.

src/user/views/custom_verify_email_view.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Call parent method to handle verification
+        parent_response = super().post(request, *args, **kwargs)
+
+        # If parent verification was successful AND we have user info
+        if parent_response.status_code == 200 and user_id:
+
+            try:
+                user = User.objects.get(id=user_id)
+                token, created = Token.objects.get_or_create(user=user)
```
gzurowski, line 39: Nit: No need to assign `created` (unused).

gzurowski (approved):
LGTM. Please note that this change introduces some risk from a security perspective which I assume has been discussed previously. To mitigate that, we should assign some (ideally short) TTL to the incoming confirmation token (`key`).

PR #2306, opened 2025-05-11
Add User Saved Content Endpoint and Database

adamandrews1:


src/user_saved/models.py
```diff
@@ -0,0 +1,75 @@
...
+                name="idx_user_list_removed",
+            ),
+            models.Index(
+                fields=["created_by", "is_removed"],
+                name="idx_user_removed",
+            ),
+        ]
+
+
+# This models a row for a user's saved content, which is a UnifiedDocument
```
gzurowski, line 39: Nit: Make a class-level Python comment instead (`""" [...] """`).

gzurowski (approved):
LGTM 👍

src/user_saved/views.py
```diff
@@ -0,0 +1,214 @@
...
+    ChangeDocumentSerializer,
+    CreateListSerializer,
+    DeleteListSerializer,
+    UserSavedListSerializer,
+)
+
+LIST_NOT_FOUND = "List not found"
+
+
+class UserSavedView(APIView):
```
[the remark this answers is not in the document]
adamandrews1: Extracted serializer creation to improve readability

src/user_saved/views.py
```diff
@@ -0,0 +1,224 @@
...
+        u_doc_id = request.query_params.get("u_doc_id")
+        paper_id = request.query_params.get("paper_id")
+        all_flag = request.query_params.get("all_items")
+        if all_flag:
+            docs = ResearchhubUnifiedDocument.objects.filter(
+                usersavedentry__created_by=request.user,
+                usersavedentry__is_removed=False,
+            ).annotate(count=Count("usersavedentry"))
+            res = {str(doc.id): doc.count for doc in docs}
+            return Response(res, status=status.HTTP_200_OK)
```
gzurowski, line 43: Shouldn't this be a separate endpoint? The shape of the data is totally different, so adding the querystring is not filtering, but changing the response type.

PR #2313, opened 2025-05-14
Add publish paper endpoint for moderators to call

koutst:
- Add publish paper endpoint
- Restrict journal publishing to moderators only
- Create a new paper and


gzurowski (approved):
LGTM 👍 

Would it be worth extracting the business logic into a service implementation to clear up the view? Not sure if it's worth it in this case, since we're adding to an existing implementation that doesn't follow such structure.

PR #2323, opened 2025-05-20
Remove history table

koutst:
- The simple history table ballooned to 270 gb and is not really used or usable.


gzurowski (approved):
LGTM 🎉

PR #2338, opened 2025-05-23
fix: Avoid removal and addition of all hubs when updating papers

gzurowski:
The previous implementation worked with different datatypes (list of IDs and list of hubs) and therefore always created a complete delta, resulting in the removal and addition of all hubs.


koutst (approved):
LGTM but is it possible to add any tests? Or are there existing tests that cover?

PR #2352, opened 2025-06-03
Grant application

koutst:
Implement grant application feature for preregistration posts
- Add GrantApplication model linking grants to preregistration posts
- Create POST /api/grant/{id}/apply endpoint for submitting applications
- Add grant_id filtering and sorting options to funding feed API
- Prevent duplicate applications with unique constraint


src/purchase/views/grant_view.py
```diff
@@ -156,3 +158,35 @@ def reopen(self, request, *args, **kwargs):
         context = self.get_serializer_context()
         serializer = self.get_serializer(grant, context=context)
         return Response(serializer.data)
+
+    @action(detail=True, methods=["post"], permission_classes=[IsAuthenticated])
+    def apply(self, request, pk=None):
```
gzurowski, line 163:
```suggestion
    def application(self, request, pk=None):
```

src/purchase/views/grant_view.py
```diff
@@ -156,3 +158,35 @@ def reopen(self, request, *args, **kwargs):
...
+            return Response({"error": "Invalid preregistration post"}, status=400)
+
+        # Check if grant is still active
+        if not grant.is_active():
+            return Response(
+                {"error": "Grant is no longer accepting applications"}, status=400
+            )
+
+        # Create application
+        application, created = GrantApplication.objects.get_or_create(
```
gzurowski, line 185: Not: Is `application` unused?

PR #2353, opened 2025-06-03
Add update grant

koutst:
- Allow grant update on post update
- Check user is moderator before creating grant


gzurowski (approved):
LGTM 👍

PR #2369, opened 2025-06-05
Reduced upvotes from 5 to 3 to trigger create_automated_bounty(item)

jeffreykoury:
Editors requested we reduce the upvote to trigger a bounty from 5 to 3 in order to ensure we are not missing any bounties due to not enough upvotes.


gzurowski (approved):
LGTM 👍

PR #2383, opened 2025-06-13
Use transaction logs instead of transaction to support smart wallets

koutst:
- Look at logs of transaction instead of top level transaction to support smart wallets


gzurowski (approved):
LGTM 👍

PR #2391, opened 2025-06-16
Add non journal filter to journal feed

koutst:
- Add a filter for papers published on RH but not in the journal

### DIFF
--- a/src/feed/views/journal_feed_view.py
+++ b/src/feed/views/journal_feed_view.py
@@ -1,7 +1,7 @@
 """
 Specialized feed view focused on papers published in the ResearchHub journal.
 This view returns papers based on their PaperVersion journal status and
-allows filtering by publication status (PREPRINT or PUBLISHED).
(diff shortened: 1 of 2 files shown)


src/feed/views/journal_feed_view.py
```diff
@@ -27,9 +28,11 @@ class JournalFeedViewSet(BaseFeedView):
     Query Parameters:
     - publication_status: Filter by publication status
       Options:
-        - PREPRINT: Only show preprints
-        - PUBLISHED: Only show published papers
-        - ALL: Show all papers (default)
+        - PREPRINT: Only show preprints in ResearchHub journal
+        - PUBLISHED: Only show published papers in ResearchHub journal
+        - NON_JOURNAL: Only show papers that are submitted but not published to
```
gzurowski, line 33: The name for this filter parameter is confusion. Would it be clearer to call it `SUBMITTED` or `NON_PUBLISHED` instead?

gzurowski (approved):

PR #2414, opened 2025-06-24
Add locking mechanism on balance to be used for funding

koutst:
- is_locked and lock_type fields on Balance model 
 - create_locked_balance() class method for easy locked balance creation
  - get_balance(include_locked=False) - excludes locked funds by default
  - Helper methods: get_available_balance() and get_locked_balance()
  - Database migration generated


src/purchase/related_models/balance_model.py
```diff
@@ -5,20 +5,57 @@
...
+        cls, user, amount, lock_type="FUNDRAISE_CONTRIBUTION", source=None
+    ):
+        """Create a locked balance entry for the user using the distributor"""
+        from django.utils import timezone
+
+        from reputation.distributions import create_locked_balance_distribution
+        from reputation.distributor import Distributor
+
+        # Create distribution for locked balance
+        distribution = create_locked_balance_distribution(amount)
```
koutst, line 43: Put in transaction

PR #2436, opened 2025-07-03
Add shell alias for Django manage

gzurowski:
Add a shell alias to execute Django's manage command more easily.

Example:

Instead of
```
cd src
python manage.py test user.tests
```

Use:
```
dj test user.tests
```


koutst (approved):

koutst:
Should we update the readme?

PR #2455, opened 2025-07-08
Add referral code assign endpoint

koutst:
- Add new referral code add_referral_code endpoint to work around issue of not being able to assign referral code in google auth

Only look at top commit

### DIFF
--- a/src/referral/views.py
+++ b/src/referral/views.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 
 from django.db.models import Count, Q, Sum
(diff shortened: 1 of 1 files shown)


src/referral/views.py
```diff
@@ -160,3 +165,90 @@ def list(self, request):
...
+                {
+                    "message": "Referral code successfully added.",
+                    "referral_signup": response_serializer.data,
+                },
+                status=status.HTTP_201_CREATED,
+            )
+
+        except Exception as e:
+            return Response(
+                {"detail": str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
```
gzurowski, line 253: This might expose internal details to the caller. Log away and return a more generic message.

PR #2502, opened 2025-08-05
Make password reset URL clickable in email template

nicktytarenko:
## What?
- Currently, the password reset URL in the email template is displayed as plain text, requiring users to copy and paste it into their browser. 
- This change makes the URL clickable for better user experience.

## How?
- Wrapped `{{ reset_url }}` in an `<a>` tag to make it clickable


gzurowski (approved):
LGTM

PR #2503, opened 2025-08-05
Feat/burn tokens from hot wallet

nicktytarenko:


src/reputation/tasks.py
```diff
@@ -510,3 +515,177 @@ def recalc_hot_score_for_open_bounties():
...
+        log_info(f"Burning transaction submitted: {tx_hash}")
+        return tx_hash
+
+    except Exception as e:
+        log_error(e, f"Failed to burn {amount} RSC from hot wallet")
+        raise
+
+
+@app.task(queue=QUEUE_PURCHASES)
+def check_hot_wallet_health():
```
koutst, line 641: How is this different than check_hot_wallet_balance?

src/reputation/tasks.py
```diff
@@ -510,3 +515,177 @@ def recalc_hot_score_for_open_bounties():
...
+    except Exception as e:
+        log_error(e, f"Failed to burn revenue RSC on {network}")
+        raise
+
+
+def _zero_out_revenue_account(revenue_account, amount):
+    """
+    Creates negative balance records to zero out the revenue account.
+    """
+    from reputation.distributions import Distribution
```
koutst, line 569: Move imports to top of file if possible

src/reputation/tasks.py
```diff
@@ -510,3 +515,177 @@ def recalc_hot_score_for_open_bounties():
 
     for bounty in open_bounties:
         bounty.unified_document.calculate_hot_score(should_save=True)
+
+
+NULL_ADDRESS = "0x0000000000000000000000000000000000000000"
+
+
+@app.task(queue=QUEUE_PURCHASES)
+def burn_revenue_rsc(network="ETHEREUM"):
```
koutst, line 524: Should we default to BASE since it's cheaper and I assume we will want to burn on Base?

koutst, line 524: I know we do a very bad job of doing this in this file but could we maybe create a wallet service within the src/reputation app that contains this logic? Like a src/reputation/services/wallet.py file.

src/reputation/tasks.py
nicktytarenko: ohhh, this is not needed. will remove it

gzurowski (commented):
The structure of the code looks very good. Wrapping the process into an (atomic) transaction is probably needed to make it more robust.

koutst (approved):

PR #2552, opened 2025-08-22
Add a review period for bounties with notifications

truelyeth:
- Add `REVIEW_PERIOD` status and `review_period_days` field to bounty model
- Transition bounties to review period instead of immediate expiration
- Add `BOUNTY_REVIEW_PERIOD_STARTED` and `BOUNTY_REVIEW_PERIOD_ENDING_SOON` notification types
- Send notifications when review period starts (to creator and submitters)
- Send 24-hour warning notification if bounty not yet awarded
- Allow bounty actions during review period
- Add tests for review period functionality
- Tested locally and works as expected


src/reputation/related_models/bounty.py
```diff
@@ -37,17 +37,23 @@ class Type(models.TextChoices):
...
         (CANCELLED, CANCELLED),
         (EXPIRED, EXPIRED),
+        (REVIEW_PERIOD, REVIEW_PERIOD),
     )
 
     expiration_date = models.DateTimeField(
         null=True,
         default=get_default_expiration_date,  # Can be null for author claim bounties
     )
+    review_period_days = models.IntegerField(
```
yattias, line 53: Rather than storing the review period time in days for each and every bounty, let's make it configurable via a constant somewhere such as `settings.py`. Unless I'm missing something I think it would be best to store the date of the review period in the DB so it is more clear when it ends

PR #2561, opened 2025-08-26
Use absolute paths for setup.py

TylerDiorio:
### Issue: `python src/manage.py setup` was failing because of relative paths

Screenshot 2025-08-26 at 2 02 48 PM

### DIFF
--- a/src/user/management/commands/setup.py
+++ b/src/user/management/commands/setup.py
@@ -2,6 +2,7 @@ import os
 
 import pandas as pd
 from allauth.socialaccount.models import SocialApp
+from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.management.base import BaseCommand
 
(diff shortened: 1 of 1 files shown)


koutst:
Hm, I think this will break if you want to run python manage.py setup from the /src directory. I think we need to construct the path using the base directory in order to make it consistent no matter where you're calling it from `os.path.join(settings.BASE_DIR, "..", "misc", "hub_hub.csv")`

TylerDiorio:
> Hm, I think this will break if you want to run python manage.py setup from the /src directory. I think we need to construct the path using the base directory in order to make it consistent no matter where you're calling it from `os.path.join(settings.BASE_DIR, "..", "misc", "hub_hub.csv")`

Good catch, I just updated it to construct using the join with BASE_DIR like you mentioned – also tested it from 2 different directories, and looks robust.
Screenshot 2025-08-26 at 2 32 58 PM

koutst (approved):

PR #2581, opened 2025-09-09
Add paper ingestion adapters

koutst:
- Initial base client and biorxiv client
- Initial base mapper and biorxiv mapper


gzurowski (approved):
LGTM 👍

PR #2590, opened 2025-09-11
Preliminary foundation for taxonomy mappings

yattias:
1. Mappings fro biorxiv, arxiv, medrxiv, chemrxiv
2. Seed script to seed DB with mapped hubs if they do not exist


koutst:
Don't we want to set the namespace to SUBCATEGORY or something similar to make filtering easier?

yattias:
Yeah, right now it is fetching the first hub in case multiples are found. 
- The tricky part about slugs is that they may be environment dependent (not a big deal though). Would the approach you had in mind:
1. Script to fetch all hubs including duplicates associated with current mapping. This would mainly be to identify hubs and their slugs
2. Update mappings to use the slugs identified by the script
3. Seed script would still be used but utilize slugs instead
 


> Hubs are unique by slug, I'm wondering if we should convert the mappings files to slugs instead of names and then look up by slugs instead. Later when we're processing papers we will be using these mappings and if we use names then there isn't a guarantee of uniqueness. Someone can come along and create a hub with the same name in the future.

koutst:
I think you would just need to update the mapping files to use slugs instead of names for the subcategory and then update the script to look up by slug instead of name. You can probably get cursor to write a script to automatically update names to slugs in the mapping files.

PR #2607, opened 2025-09-17
Add hot_score sorting for following feed

koutst:
- Add hot_score sorting for following feed
- Future improvement: Use sort_by on front page also

### DIFF
--- a/src/feed/views/feed_view.py
+++ b/src/feed/views/feed_view.py
@@ -117,14 +117,13 @@ class FeedViewSet(FeedViewMixin, ModelViewSet):
         if feed_view == "popular" or (
             feed_view == "following" and sort_by == "hot_score"
         ):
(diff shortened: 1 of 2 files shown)


src/feed/views/feed_view.py
```diff
@@ -109,14 +113,19 @@ def get_queryset(self):
...
-                    self._paper_content_type,
-                    self._post_content_type,
-                ]
-            )
+        # Handle both popular view and following view with hot_score sorting
+        if feed_view == "popular" or (
+            feed_view == "following" and sort_by == "hot_score"
+        ):
+            # Only show paper and post (for popular view only)
+            if feed_view == "popular":
```
yattias, line 121: So if `feed_view` is "following" we show other content types too? What would these be?

yattias (approved):

PR #2635, opened 2025-09-28
Fix Missing Author Field in Feed View

nicktytarenko:
What?
====
- This PR fixes a bug where the author field was missing from the main feed API (`/api/feed/`) while working correctly in the funding feed. 
- The issue is in `src/feed/signals/document_signals.py` where `create_feed_entry` calls are missing the `user_id` parameter

How?
====
- Updated the document signals to pass the `user_id` parameter


gzurowski (approved):
LGTM 👍 

I pushed up two commits to resolve the test failures.

PR #2643, opened 2025-09-30
Coinbase cors update

michaelcanova:
### Client IP Verification
We've added client IP tracking to meet Coinbase's security requirements. Now all session token requests include the user's IP address for better fraud detection.

**Changes made:**
- Updated `src/purchase/services/coinbase_service.py` to include clientIp in API calls
- Modified `src/purchase/views/coinbase_view.py` to extract and validate IP addresses

### CORS Security Improvements
Tightened up our CORS policy to only allow requests from verified ResearchHub domains. This prevents unauthorized sites from making requests to our Coinbase endpoints.

**New files:**
- `src/purchase/coinbase_cors_decorators.py` - Custom decorator for CORS validation

**Updated files:**
- `src/purchase/views/coinbase_view.py` - Now uses the new security decorator

All unauthorized access attempts are logged for monitoring purposes.


src/purchase/coinbase_cors_decorators.py
```diff
@@ -0,0 +1,103 @@
...
+def create_forbidden_response(message: str) -> JsonResponse:
+    return JsonResponse({"error": message}, status=status.HTTP_403_FORBIDDEN)
+
+
+def add_cors_headers_to_response(response, origin: str):
+    response["Access-Control-Allow-Origin"] = origin
+    return response
+
+
+def secure_coinbase_cors(view_func):
```
koutst, line 58: I think we can move this stuff into the coinbase service or some other file within a nested coinbase services folder. I think a lot of these functions should be marked private (prepended with _)

yattias: @koutst are you thinking `purchase/services/coinbase_service/` and have two files within it:

- `coinbase_service.py` 
- `coinbase_cors_decorators.py` 

As to the comment about private methods, are you suggesting the decorator method names be private or the methods within the service (e.g. `generate_jwt_token` ) ?

PR #2654, opened 2025-10-03
Script to bulk remove hubs 

yattias:
- Script to bulk remove hubs with some sophistication such as `--namespace` to remove all hubs within a namespace (e.g. "journal"
- Will skip all journals in `DEFAULT_JOURNAL_EXCLUSIONS`
- Can be used to delete any hub in the future (e.g. ones with "test" in the name, etc..)


src/hub/management/commands/find-duplicate-hubs.py
yattias, line 1: Changes in this file already applied to prod.

PR #3301, opened 2026-04-14
Capture SES message ID when sending emails

gzurowski:
Capture the SES message ID when sending Expert Finder emails.

### DIFF
--- a/src/research_ai/tasks.py
+++ b/src/research_ai/tasks.py
@@ -372,7 +372,7 @@ def send_queued_emails_task(
             failed += 1
             continue
         try:
-            send_plain_email(
+            ses_message_id = send_plain_email(
                 [rec.expert_email],
                 rec.email_subject,
                 rec.email_body,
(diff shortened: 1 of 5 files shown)


src/research_ai/tasks.py
```diff
@@ -382,6 +382,7 @@ def send_queued_emails_task(
             )
             GeneratedEmail.objects.filter(id=rec.id).update(
                 status=GeneratedEmail.Status.SENT,
+                ses_message_id=ses_message_id or "",
```
nicktytarenko, line 385: should we log something if ther eis no ses_message_id

nicktytarenko (approved):
LGTM!

src/research_ai/tasks.py
gzurowski: It's not needed. We are always sending via SES, so it will always be there. We can query the database for message IDs.

PR #3312, opened 2026-04-15
3 - (ai_peer_review): prompts + proposal review scoring

nicktytarenko:
## What?
- pure scoring/normalization logic
- add `src/ai_peer_review/prompts/` (system/user prompt templates + `proposal_review_prompts` / `rfp_summary_prompts` loaders) and tests.


src/ai_peer_review/services/proposal_review_scoring.py
```diff
@@ -0,0 +1,152 @@
...
+            score_label = _cap_sub_area_for_critical_fail(
+                dim_key, sub_key, sub_obj, score_label
+            )
+            sub_obj["score"] = score_label
+            sub_area_means.append(mean_val)
+        if not sub_area_means:
+            dim_obj["overall_score"] = "Low"
+            continue
+        dim_mean = sum(sub_area_means) / len(sub_area_means)
+        dim_obj["overall_score"] = _label_from_mean(dim_mean)
```
gzurowski: This function is quite long and hard to read. Maybe add a function-level comment explaining its purpose, along with inline comments to better document it.

src/ai_peer_review/tests/test_proposal_review_scoring.py
```diff
@@ -0,0 +1,168 @@
...
+            "novelty": {"overall_score": "Low"},
+            "impact": {"overall_score": "Low"},
+            "reproducibility": {"overall_score": "Low"},
+        }
+        rating, total = compute_overall_rating(review)
+        self.assertEqual(total, 5)
+        self.assertEqual(rating, OverallRating.POOR.value)
+
+    def test_normalize_critical_fail_caps_high(self):
+        review = {
```
gzurowski, line 88: Consider externalizing it into a separate JSON fixture file.

src/ai_peer_review/services/proposal_review_scoring.py
```diff
@@ -0,0 +1,197 @@
...
+
+SCORE_MAP = {"High": 3, "Medium": 2, "Low": 1}
+
+ANSWER_TO_NUMERIC = {
+    "yes": 1.0,
+    "partial": 0.5,
+    "no": 0.0,
+}
+
+CRITICAL_FAIL_KEYS = {
```
yattias: I think this needs some documentation. I think we need to get to a place where the code is more human readable

src/ai_peer_review/services/proposal_review_scoring.py
```diff
@@ -0,0 +1,197 @@
...
+def _answer_to_numeric(value: Optional[str]) -> Optional[float]:
+    if not isinstance(value, str):
+        return None
+    value_lower = value.strip().lower()
+    if value_lower == "n/a":
+        return None
+    return ANSWER_TO_NUMERIC.get(value_lower)
+
+
+def _cap_sub_area_for_critical_fail(
```
yattias: Still a bit hard to read. maybe documentation will help? perhaps once tyler provides the prompt, this will be easier to discern

nicktytarenko: added a short [README](https://github.com/ResearchHub/researchhub-backend/blob/9205b068410e5b8b6d1583a12b233a44a6bbb865/src/ai_peer_review/docs/proposal_review_scoring.md) explaining this. 
In brief: we don't use the LLM's headline `score` / `overall_score` as source of truth, we **recompute them from the Yes/No/Partial answers** using fixed rules _(thresholds, optional N/A handling, critical-fail caps)_.

nicktytarenko: This [test](https://github.com/ResearchHub/researchhub-backend/pull/3312/changes#diff-a200b22a67674ed4040f53427b069d0cf32ff3cf8ea033d26a12c9488a31ebdbR45-R63) shows normalization in action: the LLM can return `score: "High"`, but we recompute from the Yes/No/Partial answers and apply the critical-fail cap, so a `No` on `go_no_go_gates` under `fundability.timeline_realism` forces the stored score to **Medium**.

PR #3314, opened 2026-04-15
4 - (ai_peer_review): add external context and LLM service layer

nicktytarenko:
## Summary
- author context text, OpenAlex/ORCID-backed external researcher context
- Bedrock Converse wrapper for peer-review calls
- OpenAI web-search context helper

## Changes
- Add `services/author_context.py`: builds factual, on-platform lines (name, headline, affiliation, bio, ORCID, OpenAlex ids on profile, h/i10, education count, links)
- Add `services/researcher_external_context.py` (OpenAlex fetch + format).
- tests


gzurowski (commented):
Some of the functions in this PR are extremely long and hard to read. I wonder if this can be improved?

nicktytarenko:
> Some of the functions in this PR are extremely long and hard to read. I wonder if this can be improved?

Fair point on the length. I think you're referring to `format_openalex_author_record`. This is mostly straightforward "read field -> append one line" mappings for the prompt.
I kept it in one place intentionally, so it's easy to see exactly what we expose compared to the raw OpenAlex payload.
There's a short note in the code plus a link to the OpenAlex author schema for reference.

Happy to split it into smaller helpers if you think that would improve readability, but I don't think the logic itself is particularly complex.

src/ai_peer_review/services/openai_web_context_service.py
```diff
@@ -0,0 +1,151 @@
...
+logger = logging.getLogger(__name__)
+
+OPENAI_WEB_CONTEXT_MODEL = getattr(
+    settings,
+    "AI_PEER_REVIEW_OPENAI_WEB_CONTEXT_MODEL",
+    "gpt-5.4-mini",
+)
+
+
+class OpenAIWebContextService:
```
yattias: I wonder if this should be called something like OpenAIReviewContext. The part that's missing is the review

yattias (approved):

PR #3317, opened 2026-04-16
5 - (ai_peer_review): Proposal review pipeline + Celery

nicktytarenko:
## What?

- Proposal reviews can run end-to-end in the background: 
  - load proposal, 
  - enrich with researcher/public-record, 
  - call the LLM, 
  - normalize scores and ratings;
- Celery tasks exist so that work can be queued instead of running inline; 
- there is logic to build a funder-facing comparison across completed proposal reviews for a grant when that flow is invoked.
- and grant text can be summarized


src/ai_peer_review/tasks.py
```diff
@@ -0,0 +1,15 @@
+from ai_peer_review.services.proposal_review_service import run_proposal_review
+from ai_peer_review.services.rfp_summary_service import run_rfp_summary
+from researchhub.celery import app
+
+
+@app.task
+def process_proposal_review_task(review_id: int):
```
yattias, line 7: What would trigger this?

src/ai_peer_review/services/researcher_external_context.py
```diff
@@ -176,3 +181,145 @@ def build_researcher_external_context_for_author(
         openalex_author_ref=oa_ref,
         client=client,
     )
+
+
+def fetch_orcid_works(
```
yattias, line 186: I think the orcid stuff is okay here but next time, for the sake of keeping PRs smaller. We could add this sort of thing on top iteratively

yattias (approved):

src/ai_peer_review/tasks.py
nicktytarenko: There is still no automation to trigger this on every proposal creation or update. Instead there is an API endpoint (**POST** `/api/ai_peer_review/proposal-review/` that is available for mods/editors only) that starts the review workflow _(the API returns 202 and the review is queued for processing)_.

PR #3335, opened 2026-04-20
Remove the DOI creation upon publishing from Notebook

michaelcanova:
## Overview ##

This PR removes the portion where a DOI is created when anything in the notebook is published.

### DIFF
--- a/src/researchhub_document/views/researchhub_post_views.py
+++ b/src/researchhub_document/views/researchhub_post_views.py
@@ -340,7 +340,6 @@ class ResearchhubPostViewSet(ReactionViewActionMixin, ModelViewSet):
                     status=400,
                 )
 
-            created_by = request.user
             hubs = data.get("hubs", None)
             renderable_text = data.get("renderable_text", "")
             title = data.get("title", "")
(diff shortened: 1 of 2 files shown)


src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -126,7 +124,6 @@ def create_researchhub_post(self, request):
         document_type = data.get("document_type")
         editor_type = data.get("editor_type")
         title = data.get("title", "")
-        assign_doi = data.get("assign_doi", False)
```
gzurowski, line 129: I was wondering if it makes sense to retain this flag, unless we really want to assign DOIs to every post created. Or are we totally dropping this optionality and always seek to create DOIs?

gzurowski (approved):

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: As we discussed, this pertains to just items in the notebook (grants, preregistrations, and posts).  And even then, grants go through their own process through moderation approval.  In the current scope of things, since we are no longer assigning DOIs to anything in the notebook upon publication, there's really no need for the flag anymore.

PR #3336, opened 2026-04-20
Task populates DOIs for Notebook Posts

michaelcanova:
## Overview ##

Part 2 of  https://github.com/ResearchHub/researchhub-backend/pull/3335

Now that DOIs are not created when a work is published, we need a task.  This task runs daily, and checks for any proposals that are older than 7 days, and assigns them a DOI.  This 7 day grace period allows us to detect and remove spam works before assigning them DOIs..  It will not assign DOI's to both removed and flagged content.

### DIFF
--- a/src/researchhub_document/tasks.py
+++ b/src/researchhub_document/tasks.py
@@ -42,14 +42,12 @@ def assign_post_dois():
                 post.save(update_fields=["doi"])
                 assigned_count += 1
             else:
-                sentry.log_error(
-                    RuntimeError(
-                        f"Crossref API failure for post {post.id}: "
-                        f"status {response.status_code}"
-                    )
+                logger.error(
+                    f"Crossref API failure for post {post.id}: "
+                    f"status {response.status_code}"
                 )
-        except Exception as error:
-            sentry.log_error(error)
+        except Exception:
+            logger.exception(f"Failed to assign DOI to post {post.id}")
 
     logger.info(f"Assigned DOIs to {assigned_count}/{total} eligible posts")
 
--- a/src/researchhub_document/tasks.py
+++ b/src/researchhub_document/tasks.py
@@ -43,7 +43,7 @@ def assign_preregistration_dois():
                 assigned_count += 1
             else:
                 sentry.log_error(
-                    Exception(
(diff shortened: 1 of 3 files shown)


src/researchhub_document/tasks.py
```diff
@@ -1,7 +1,57 @@
...
+            doi = DOI()
+            author = post.created_by.author_profile
+            response = doi.register_doi_for_post([author], post.title, post)
+
+            if response.status_code == 200:
+                post.doi = doi.doi
+                post.save(update_fields=["doi"])
+                assigned_count += 1
+            else:
+                sentry.log_error(
```
gzurowski: Use regular logging, not the Sentry logger.

src/researchhub_document/tasks.py
```diff
@@ -1,7 +1,57 @@
...
+                    RuntimeError(
+                        f"Crossref API failure for post {post.id}: "
+                        f"status {response.status_code}"
+                    )
+                )
+        except Exception as error:
+            sentry.log_error(error)
```
gzurowski: Use regular logging, not the Sentry logger.

src/researchhub_document/tasks.py
```diff
@@ -1,7 +1,55 @@
...
+    eligible_posts = (
+        ResearchhubPost.objects.filter(
+            doi__isnull=True,
+            created_date__lte=week_ago,
+            unified_document__is_removed=False,
+            flags__isnull=True,
+        )
+        .exclude(document_type=GRANT)
+        .select_related("created_by__author_profile", "unified_document")
+    )
```
gzurowski, line 32: Is this the correct queryset to target posts for the creation of DOIs? It seems it will assign DOIs to every post created within a week.

michaelcanova: That makes sense, it could possibly grab question types as well, so I updated it to only target posts and preregistrations.  thank you!

gzurowski: I think this makes sense, although I am still wondering why we would assign DOIs to discussion posts?

gzurowski (approved):

PR #3364, opened 2026-04-24
[expert-finder] 3 - JSON system prompt, helpers, v2 run + Celery task, excluded search IDs

nicktytarenko:
What?
=====
- `expert_finder_system_v2.txt` defines the v2 **output JSON shape**
- We’re creating a **parallel path#2** alongside the existing Expert Finder flow _(not wired to any endpoint yet)_
  - Added a separate `ExpertFinderServiceV2.process_expert_search_v2` to use the new prompt and parsing logic


src/research_ai/services/expert_finder_v2.py
```diff
@@ -0,0 +1,418 @@
...
+from research_ai.services.progress_service import TaskType
+from research_ai.services.report_generator_service import (
+    generate_csv_file_v2,
+    generate_pdf_report_v2,
+    upload_report_to_storage,
+)
+
+logger = logging.getLogger(__name__)
+
+V2_PROMPT_EXPERT_RESERVE_PCT = 0.1
```
koutst: Maybe add a comment here for this magic number here, I know there's one in _prompt_expert_count_for_round but that could move and is indirect

src/research_ai/prompts/expert_finder_prompts.py
koutst: I second the sonarqube comment `Define a constant instead of duplicating this literal "All States" 4 times.`

src/research_ai/services/expert_finder_v2.py
```diff
@@ -0,0 +1,418 @@
...
+                        continue
+                    if search_id_emails and em in search_id_emails:
+                        continue
+                    kept.append(row)
+                if round_num == 1 and search_id_emails and n_before > 0 and not kept:
+                    all_filtered_by_exclusion = True
+
+                batch = [r for r in kept if not self._expert_row_suggests_deceased(r)]
+
+                batch = self._dedupe_experts_by_normalized_email(batch)
```
koutst, line 307: Dedupe is happening within the for loop, is it possible this will cause repeats across batches?

koutst (approved):

PR #3365, opened 2026-04-24
Add use_credits param to fundraise contribution

koutst:
- If use_credits true, only use funding credits, if false only use balance

### DIFF
--- a/src/purchase/tests/test_fundraise_service.py
+++ b/src/purchase/tests/test_fundraise_service.py
@@ -359,7 +359,7 @@ class CloseFundraiseTests(TestCase):
 
         contribution_amount = Decimal("100")
         purchase, error = self.fundraise_service.create_rsc_contribution(
-            contributor, self.fundraise, contribution_amount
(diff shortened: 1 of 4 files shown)


src/purchase/tests/test_fundraise_service.py
```diff
@@ -590,6 +574,117 @@ def test_close_fundraise_all_unlocked_stays_unlocked(self):
...
+        )
+        Balance.objects.create(
+            amount=500, user=contributor, content_type=dist_ct, is_locked=False
+        )
+
+        purchase, error = self.fundraise_service.create_rsc_contribution(
+            contributor, self.fundraise, Decimal("100"), use_credits=True
+        )
+        self.assertIsNone(purchase)
+        self.assertEqual(error, "Insufficient funding credits")
```
yattias, line 632: Similar to the assertion in previous test: 
```
# Unlocked balance is untouched.
```

can we add an assertion that funding credits are untouched?

yattias (approved):

PR #3366, opened 2026-04-24
Update OpenSearch upon item removal from Moderation Dashboard

michaelcanova:
## Overview ##

This PR adds functionality to clear OpenSearch when items are removed from the Moderation Dashboard
    1. If an individual item is removed (Remove button), it will remove that item
    2. If a user is removed (Remove + Suspend) all items pertaining to that user will be removed


src/discussion/tests/test_censor.py
```diff
@@ -0,0 +1,20 @@
...
+from unittest.mock import Mock, patch
+
+from researchhub_document.models import ResearchhubUnifiedDocument
+
+
+class TestCensorUnifiedDocument(TestCase):
+
+    @patch("discussion.views.remove_from_search_index")
+    def test_resolves_unified_document_to_inner_doc(self, mock_remove):
+        from discussion.views import censor
```
gzurowski, line 11: Push to top

src/discussion/tests/test_censor.py
```diff
@@ -0,0 +1,20 @@
+from unittest import TestCase
+from unittest.mock import Mock, patch
+
+from researchhub_document.models import ResearchhubUnifiedDocument
+
+
+class TestCensorUnifiedDocument(TestCase):
+
+    @patch("discussion.views.remove_from_search_index")
+    def test_resolves_unified_document_to_inner_doc(self, mock_remove):
```
gzurowski, line 10: Nit: Add AAA markers

src/search/tasks.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    from paper.models import Paper
+    from researchhub_document.related_models.researchhub_post_model import (
+        ResearchhubPost,
+    )
+
+    sync_search_index(Paper.objects.filter(is_removed=True))
+    sync_search_index(
+        ResearchhubPost.objects.filter(unified_document__is_removed=True)
+    )
```
gzurowski, line 24: This queryset will grow large over time which will cause large amounts of registry updates redundantly.

gzurowski (commented):
The change couples discussion, hub, paper, RH document with the search app that already depends on those, creating a dependency circle between these modules. I would prefer a solution that doesn't introduce this form of coupling.

PR #3370, opened 2026-04-25
3 - ai-review/prompt-update

nicktytarenko:
What?
=====
- Adds `proposal_review_comment_service` to create/update comment on the Proposal's document, with a fixed AI author user
- convert result_data to TipTap json for saving it to the comment table


src/ai_peer_review/services/proposal_review_comment_service.py
```diff
@@ -0,0 +1,276 @@
...
+
+from ai_peer_review.models import ProposalReview, ReviewStatus
+from researchhub_comment.constants.rh_comment_content_types import TIPTAP
+from researchhub_comment.models import RhCommentModel, RhCommentThreadModel
+from review.models import Review
+from user.models import User
+
+logger = logging.getLogger(__name__)
+
+AI_EXPERT_EMAIL = "ai-review@researchhub.foundation"
```
yattias, line 13: This will be env variable in future?

yattias (approved):

src/ai_peer_review/services/proposal_review_comment_service.py
nicktytarenko: Ohh, we can do this 👍🏼
I noticed we have some RHF accounts hardcoded as constants, and I followed the same approach for the AI review:
https://github.com/ResearchHub/researchhub-backend/blob/main/src/user/related_models/user_model.py#L25-L26

But it should be easy to move this into settings instead 👌🏼

PR #3373, opened 2026-04-26
5 - ai-review: add proposal key insights service  + management command

nicktytarenko:
What?
=====

- Implements `proposal_key_insights_service` to gather all the data needed to run `run_proposal_key_insights`
- Adds a management command to execute `run_proposal_key_insights`


src/ai_peer_review/management/commands/run_proposal_key_insights.py
```diff
@@ -0,0 +1,212 @@
+"""
+Run the key-insights LLM pass (TLDR + pros/cons) for completed proposal reviews.
```
yattias: Is this "for completed proposal reviews." or for "funding opportunity + proposal + reviews". It is nuanced but the text makes me think this is review only

src/ai_peer_review/management/commands/run_proposal_key_insights.py
```diff
@@ -0,0 +1,212 @@
...
+
+        reviews = list(
+            reviews_qs.select_related("unified_document", "grant").order_by("id")
+        )
+        total = len(reviews)
+        generated = 0
+        skipped = 0
+        failed = 0
+
+        completed_before_ids = set(
```
yattias: the "completed" status threw me off a bit. I think what it just means is that the proposal review has been "processed" successfully which is the case with 100% of our reviews on prod. The reason I was thrown off is because "completed" makes it feel like it is a distinct phase reviews go through that is significantly different and time-separated but in actuality it just implied processed. 

TLDR; If my assumption above is correct, let's rename "processed_reviews"

src/ai_peer_review/services/proposal_review_comment_service.py
```diff
@@ -238,6 +238,38 @@ def _review_thread_reference(review: ProposalReview) -> str:
...
+    The AI proposal review comment.
+    """
+    post = review.unified_document.posts.first()
+    if post is None:
+        return None
+    ai_user = get_ai_expert_user()
+    if ai_user is None:
+        return None
+    content_type = ContentType.objects.get_for_model(post)
+    thread = RhCommentThreadModel.objects.filter(
```
yattias, line 249: I think this model is legacy and while at it, would be good if you mark it as deprecated with a comment. Last record in the DB from this model is from: 2025-10-25 14:46:09.559899+00

src/ai_peer_review/services/proposal_key_insights_service.py
```diff
@@ -0,0 +1,231 @@
...
+    ProposalReview,
+    ReviewStatus,
+)
+from ai_peer_review.prompts.proposal_review_prompts import (
+    build_proposal_key_insights_user_prompt,
+    get_proposal_key_insights_system_prompt,
+)
+from ai_peer_review.services.bedrock_llm_service import BedrockLLMService
+from ai_peer_review.services.proposal_review_comment_service import (
+    get_ai_expert_user,
```
yattias: I haven't looked into the implementation of this function but if it works similarly to other accounts we have, we should bind a class method to `User.objects` 

Such as:

```
User.objects.get_community_revenue_account()
```

src/ai_peer_review/management/commands/run_proposal_key_insights.py
yattias: furthermore, we have completed on key insights in addition to reviews which was adding to my confusion

src/ai_peer_review/management/commands/run_proposal_key_insights.py
```diff
@@ -0,0 +1,212 @@
...
+                continue
+
+            if ki.status == ReviewStatus.FAILED:
+                failed += 1
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"[{i}/{total}] {label} FAILED: {ki.error_message[:200]!r}"
+                    )
+                )
+            elif not force and before and ki.status == ReviewStatus.COMPLETED:
```
yattias: I think I was confused here because we are comparing key insight status to review status. I know they resolve to the same string but when looking at a glance I think we are looking at reviews and not key insights

src/ai_peer_review/management/commands/run_proposal_key_insights.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+        if force:
+            self.stdout.write(
+                self.style.NOTICE(
+                    "--force: re-running key insights that were already completed."
+                )
+            )
+
+        for i, review in enumerate(reviews, start=1):
+            before = review.id in completed_before_ids
```
yattias:
```
before = review.id in completed_before_ids
```

At a glance, this makes it look like `before` is a review and not a key insight. That is because earlier on, we collect the `values_list("proposal_review_id", flat=True)`. I think we should update the var name to indicate what is key insight VS review so reader can quickly skim

src/ai_peer_review/services/proposal_key_insights_service.py
```diff
@@ -0,0 +1,231 @@
...
+            )
+        )
+    return out
+
+
+def run_proposal_key_insights(
+    review_id: int, *, force: bool = False
+) -> ProposalKeyInsight:
+    """
+    Run the key-insights Bedrock pass for a completed ``ProposalReview`` and
```
yattias: I think we should update the comment here to make it easy for human + AI to know what this function does more specifically which is the generation of key insights through the processing of proposal, rfp, human reviews, ai review,

src/ai_peer_review/services/proposal_key_insights_service.py
yattias, line 1: Can we make the service class based to adhere with our convention of services? If there is a reason against it, LMK

src/ai_peer_review/services/proposal_review_comment_service.py
```diff
@@ -238,6 +238,38 @@ def _review_thread_reference(review: ProposalReview) -> str:
...
+        content_type=content_type,
+        object_id=post.id,
+        thread_type=COMMUNITY_REVIEW,
+        thread_reference=_review_thread_reference(review),
+    ).first()
+    if thread is None:
+        return None
+    return (
+        RhCommentModel.objects.filter(
```
yattias, line 258: I'm not seeing a place in the code where we create an AI review comment. LMK where that happens

yattias (approved):
Left a few comments I think would make the code easier to read. For next time, we could make this PR even more granular. I think we could separate out the script from the service code

src/ai_peer_review/tests/test_proposal_review_comment_service.py
yattias, line 157: I think this test would need to be updated. We don't need to created thread and comment since thread is an artifact

src/ai_peer_review/services/proposal_review_comment_service.py
```diff
@@ -12,7 +13,6 @@
 
 AI_EXPERT_EMAIL = "ai-review@researchhub.foundation"
```
yattias: I wonder if we should put this in `user_model` alongside the other foundation emails

src/ai_peer_review/tests/test_proposal_review_comment_service.py
yattias, line 1: I think we need a test for ensuring key insights are generated

src/ai_peer_review/management/commands/run_proposal_key_insights.py
nicktytarenko: For each review, the model receives the proposal body, grant context, the AI peer-review text, and assessed human community reviews.
Updated the description accordingly

src/ai_peer_review/services/proposal_review_comment_service.py
nicktytarenko: I'm a bit confused 😕  
I can see comment threads in the DB. Or were you pointing to the wrong line in the codebase?
image

src/ai_peer_review/management/commands/run_proposal_key_insights.py
nicktytarenko: updated to `key_insight_was_already_done = review.id in processed_reviews`

src/ai_peer_review/services/proposal_review_comment_service.py
nicktytarenko: This is implemented in this PR:
[https://github.com/ResearchHub/researchhub-backend/pull/3394](https://github.com/ResearchHub/researchhub-backend/pull/3394)

They'll be created/updated via signals _(and there's also a management command to run this manually if needed)._

src/ai_peer_review/tests/test_proposal_review_comment_service.py
nicktytarenko: this is done in the next PR 👍🏼

PR #3385, opened 2026-04-27
Deleted comments no longer appear on work pages

michaelcanova:
## Overview ##

This PR removes the ability for deleted/removed comments to show on works pages.  A lot of code associated to showing/delivering said removed/deleted comments has been removed as well, as they are no longer needed.

Also, if a comment is deleted, all of its sub comments are also deleted.


yattias (approved):
I think we need some tests to ensure that deleted comments do not show up (unless we have some in place)

PR #3396, opened 2026-04-29
Add staking endpoints for overview page

koutst:


docs/STAKING_YIELD_API.md
yattias, line 1: Should we put docs in this folder? Do you think it's helpful? Is this mainly for helping with integration?

koutst: It was mainly for the hand off when I was taking a couple of weeks off so you could work on the front end if needed. Now I just left it because I think it can be helpful

PR #3402, opened 2026-04-30
User-generated content feed for moderators

TylerDiorio:
Adds `GET /api/feed/user_generated/` — a moderator-only feed of entries with a non-null user, so we can have a feed of all content being generated on the site.


src/feed/views/feed_view.py
```diff
@@ -162,3 +164,36 @@ def get_queryset(self):
...
+                "user__userverification",
+                "unified_document",
+            )
+            .order_by("-action_date")
+        )
+
+        page = self.paginate_queryset(queryset)
+        serializer = self.get_serializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        self.add_user_votes_to_response(request.user, response.data)
```
yattias, line 197: Let's skip this part. I think we can do it safely

src/feed/views/feed_view.py
```diff
@@ -162,3 +164,36 @@ def get_queryset(self):
         )
 
         return queryset
+
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="user_generated",
```
yattias, line 171: I don't think we need this actually

yattias (approved):

PR #3403, opened 2026-04-30
API changes for Proposal Funding Opportunities 

michaelcanova:
## Overview ##

This PR adds parameters when pulling data for a proposal's work page to show the Funding Opportunities section.

### DIFF
--- a/src/researchhub_document/serializers/researchhub_post_serializer.py
+++ b/src/researchhub_document/serializers/researchhub_post_serializer.py
@@ -236,8 +236,16 @@ class ResearchhubPostSerializer(ModelSerializer, GenericReactionSerializerMixin)
 
         ud_id = unified_document.id
         applications = list(post.grant_applications.all())
-        grant_ids = [app.grant_id for app in applications]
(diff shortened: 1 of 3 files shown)


src/researchhub_document/serializers/researchhub_post_serializer.py
```diff
@@ -243,6 +243,16 @@ def get_grants(self, post):
                 if review is not None
                 else None
             )
+
+            grant_post = None
+            grant_posts = grant.unified_document.posts.all()
```
yattias: Can you do some comparison via the django toolbar to see how many more queries this is adding?

I wonder if we can prefetch posts above and whether that would speed things up. Also wonder if we can avoid `grant.unified_document.posts.all()` and do something like ResearchhubPost.objects.get(id=grant.unified_document.id)` if it results in improved performance

michaelcanova: Good callout, updated and fixed!

yattias (approved):

PR #3414, opened 2026-05-05
Use logging in Amplitude webhook

gzurowski:
- Use logger instead of Sentry utils (error-level logs are automatically be pushed to Sentry).
- Don't reveal exception to caller (Amplitude).


yattias (approved):
Nice

PR #3434, opened 2026-05-06
Adding is_funder parameter

michaelcanova:
## Overview ##

This PR adds the is_funder parameter on the user to help the front end determine whether or not a user should see the funder dashboard 

Based on if the user has created funding opportunities that have been approved by the moderation team.

### DIFF
--- a/src/user/views/user_views.py
+++ b/src/user/views/user_views.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta
 
 from allauth.account.models import EmailAddress
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import F, Q, Sum
+from django.db.models import Exists, F, OuterRef, Q, Sum
 from django.db.models.functions import Coalesce
 from django.utils import timezone
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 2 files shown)


src/user/views/user_views.py
```diff
@@ -49,9 +50,18 @@
 
 
 class UserViewSet(FollowViewActionMixin, viewsets.ModelViewSet):
-    queryset = User.objects.select_related(
-        "userverification",
-    ).filter(is_suspended=False)
+    queryset = (
+        User.objects.select_related("userverification")
+        .annotate(
+            _is_funder=Exists(
```
yattias: For consistency, let's just call it `is_funder` since we don't use private variable convention in annotations

yattias (approved):

src/user/views/user_views.py
```diff
@@ -49,9 +50,18 @@
...
+                Grant.objects.filter(
+                    created_by=OuterRef("pk"),
+                    status__in=[Grant.OPEN, Grant.CLOSED, Grant.COMPLETED],
```
koutst: I think we can remove CLOSED

PR #3446, opened 2026-05-10
Add cached invited-experts overview endpoint

nicktytarenko:
What?
=====

- Introduces **GET** `/expert-finder/invited-experts/overview/`, same auth/permissions as other expert-finder views.
- Query filters: 
  - optional `unified_document_id`, 
  - optional `start` / `end` as calendar dates; 
- Responses cached for 1h with a stable cache key from those params.

Example: `.../api/research_ai/expert-finder/invited-experts/overview/?unified_document_id=42&start=2026-01-01`

### DIFF
--- a/src/research_ai/views/expert_finder_views.py
+++ b/src/research_ai/views/expert_finder_views.py
@@ -242,15 +242,6 @@ INVITED_CACHE_SEC = 60 * 60 * 3  # 3 hours
 INVITED_OVERVIEW_CACHE_TTL = 60 * 60 * 1  # 1 hour
 
 
-def _invited_expert_overview_cache_key(*, unified_document_id, start, end):
-    ud_part = (
-        str(int(unified_document_id)) if unified_document_id is not None else "none"
-    )
-    start_part = start.isoformat() if start is not None else "none"
-    end_part = end.isoformat() if end is not None else "none"
-    return f"invited_expert_overview:ud={ud_part}:start={start_part}:end={end_part}"
-
-
 class InvitedExpertOverviewView(APIView):
     """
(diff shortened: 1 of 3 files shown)


src/research_ai/views/expert_finder_views.py
```diff
@@ -236,6 +239,69 @@ def patch(self, request, expert_id):
...
+INVITED_OVERVIEW_CACHE_TTL = 60 * 60 * 1  # 1 hour
+
+
+def _invited_expert_overview_cache_key(*, unified_document_id, start, end):
+    ud_part = (
+        str(int(unified_document_id)) if unified_document_id is not None else "none"
+    )
+    start_part = start.isoformat() if start is not None else "none"
+    end_part = end.isoformat() if end is not None else "none"
+    return f"invited_expert_overview:ud={ud_part}:start={start_part}:end={end_part}"
```
gzurowski: Nit: This private function is only used by the view. Should it be moved into the view class?

src/research_ai/views/expert_finder_views.py
```diff
@@ -236,6 +239,69 @@ def patch(self, request, expert_id):
...
+        if cached_payload is not None:
+            return Response(cached_payload)
+
+        overview = get_invited_expert_overview(
+            unified_document_id=unified_document_id,
+            start=start,
+            end=end,
+        )
+        cached_at = timezone.now()
+        body = {**overview, "cached_at": cached_at}
```
gzurowski: This mixes metadata (cache date) with the actual payload. I suggest separating the two: either by nesting the caching date in a `meta` structure or in a custom response header (`rh-cached-at`).

gzurowski (approved):

PR #3449, opened 2026-05-12
OpenSearch Management Commands

michaelcanova:
## Overview ##

This PR adds management commands to clear works that have been removed from OpenSearch.  The commands have also been added to the README.


src/search/tests/test_remove_deleted_docs_from_index.py
```diff
@@ -0,0 +1,261 @@
...
+            call_command(COMMAND, "--index=post", stdout=StringIO())
+
+        # Assert
+        ids = _get_updated_ids(mock_update)
+        self.assertIn(self.post_removed.id, ids)
+        self.assertNotIn(self.post_active.id, ids)
+
+    # --- Hub index ---
+
+    def test_hub_index_excludes_journals(self):
```
yattias: I don't think we need to exclude journal if they are deleted. Whatever is marked as deleted we should remove. Could be a journal we don't want shown

src/search/tests/test_remove_deleted_docs_from_index.py
```diff
@@ -0,0 +1,261 @@
...
+            f"{PATCH_PREFIX}.PaperDocument.update"
+        ) as mock_update:
+            out = StringIO()
+            call_command(COMMAND, "--index=paper", stdout=out)
+
+        # Assert
+        mock_update.assert_not_called()
+        self.assertIn("0 still in index", out.getvalue())
+
+    def test_skips_opensearch_check_when_no_removed_docs_in_database(self):
```
yattias: We are trying to limit tests to only necessary ones to avoid slowing things down. This particular test can be removed because it doesn't test any core logic

src/search/tests/test_remove_deleted_docs_from_index.py
```diff
@@ -0,0 +1,261 @@
...
+        ) as mock_update:
+            call_command(COMMAND, "--index=paper", stdout=StringIO())
+
+        # Assert
+        mock_conn.assert_not_called()
+        mock_update.assert_not_called()
+
+    # --- Error handling ---
+
+    def test_update_exception_is_logged_without_aborting(self):
```
yattias: Ditto

src/search/management/commands/remove_deleted_docs_from_index.py
```diff
@@ -0,0 +1,172 @@
...
+        label: str = config["label"]
+        client = doc._get_connection()
+        opensearch_index = doc._index._name
+
+        removed_in_db = 0
+        still_in_index = 0
+        processed = 0
+        errors = 0
+
+        for batch in batched(qs.iterator(chunk_size=batch_size), batch_size):
```
yattias: can we potentially do without batch function? we have `batch_size` passed into both the iterator and `batched` function. wonder if we can only use one.

src/search/management/commands/remove_deleted_docs_from_index.py
```diff
@@ -0,0 +1,172 @@
...
+            removed_in_db += len(batch)
+            found, removed, errs = self._check_and_remove_batch(
+                doc, client, opensearch_index, label, batch, dry_run
+            )
+            still_in_index += found
+            processed += removed
+            errors += errs
+
+        self.stdout.write(
+            f"{label}: {removed_in_db} removed in database, "
```
yattias: I think `removed_in_db` is unlikely but, potentially wrong since it is the batch size and not the actual length of those removed. Perhaps `len(removed)` is better here

yattias (approved):

PR #3450, opened 2026-05-12
Add rfp forced privacy settings

koutst:
Add an option for an RFP to dictate privacy setting of preregistrations that apply


src/researchhub_document/tests/test_private_preregistration.py
```diff
@@ -436,3 +436,279 @@ def test_comment_on_public_doc_creates_feed_entry(self):
...
+        self.assertEqual(response.status_code, 200)
+        post = ResearchhubPost.objects.get(id=response.data["id"])
+        self.assertFalse(post.unified_document.is_public)
+        action = Action.objects.get(
+            content_type=ContentType.objects.get_for_model(ResearchhubPost),
+            object_id=post.id,
+        )
+        self.assertFalse(action.display)
+
+    def test_create_with_private_required_grant_explicit_public_errors(self):
```
yattias, line 531: Really small: Reading these tests could be made easier if we replace `create` with apply or `create_proposal` or `create_and_apply`.

yattias (approved):

src/researchhub_document/tests/test_private_preregistration.py
koutst: This is creating a post though, the direct apply tests are below

PR #3456, opened 2026-05-15
Expert finder invite private access and manual addition

koutst:
- Allow user to create manual expert
- Invited experts should have access to private preregistration they're invited to


nicktytarenko (approved):
LGTM

PR #3457, opened 2026-05-15
Risk Score Scaffolding

michaelcanova:
## Overview ##

This PR creates the basic scaffolding for the new risk score system.  It includes the migrations, model, constants, and basic score values.


src/risk_score/models.py
```diff
@@ -0,0 +1,128 @@
...
+        EventType.ACCOUNT_CREATED: None,
+        EventType.BACKFILL: None,
+    }
+
+    VOTE_TYPES = {
+        EventType.CONTENT_UPVOTED,
+        EventType.CONTENT_DOWNVOTED,
+    }
+
+    ONE_TIME_TYPES = {
```
yattias, line 74: Let's put a comment explaining what this is. 

In one of the future PRs, you should add some idempotent guard that prevents these one time events from being captured more than once

src/risk_score/models.py
```diff
@@ -0,0 +1,128 @@
...
+        EventType.PEER_REVIEW_ASSESSED: -5,
+        # One-time profile signals
+        EventType.EXPERT_FINDER_SIGNUP: -51,
+        EventType.EDU_EMAIL_SIGNUP: -20,
+        EventType.ORCID_VERIFIED_EDU: -10,
+        EventType.GOOGLE_SIGNUP: -10,
+        EventType.ACCOUNT_AGE_BONUS: -5,
+        EventType.PERSONA_VERIFIED_WHITELISTED: -51,
+        EventType.PERSONA_VERIFIED_NON_WHITELISTED: -10,
+        EventType.ACCOUNT_CREATED: None,
```
koutst: Why are there event types with None? Also why is backfill an event type, shouldn't a backfill just fill in the appropriate event types?

src/risk_score/constants.py
```diff
@@ -0,0 +1,8 @@
+DEFAULT_SCORE = 100
+TRUSTED_THRESHOLD = 50
+RESTRICTED_THRESHOLD = 150
+SCORE_FLOOR = 0
+SCORE_CEILING = 300
```
koutst: Why are we implementing a ceiling and a floor and why is the ceiling 200 above the start point while the floor is 100 below. That means that if a user does a bunch of good stuff (-400) to followed by a bunch of bad stuff (+200) their score will be 200 while if you flip the order their score will be 0.

src/risk_score/constants.py
```diff
@@ -0,0 +1,8 @@
...
+
+DAILY_UPVOTE_SCORE_CAP = 10
+DAILY_DOWNVOTE_SCORE_CAP = 10
```
koutst: I thought we agreed against this since it's easily gameable ?

src/risk_score/models.py
```diff
@@ -0,0 +1,128 @@
...
+from utils.models import DefaultModel
+
+
+class RiskScore(DefaultModel):
+    user = models.OneToOneField(
+        "user.User",
+        on_delete=models.CASCADE,
+        related_name="risk_score",
+    )
+    score = models.IntegerField(default=DEFAULT_SCORE, db_index=True)
```
koutst: Move index into Meta

src/risk_score/models.py
michaelcanova: Put a comment in for now, and yes the idempotent guard is in the service layer in the next PR!

src/risk_score/models.py
michaelcanova: Updated (and removed, after testing we dont need those fields at all).  Thank you for the feedback!

src/risk_score/constants.py
michaelcanova: That is a valid point, it was to make it where good actors who did bad things could get back to neutral/high risk easier, but we really dont need it, so i removed the floor and ceiling.  Thank you for the feedback!

src/risk_score/constants.py
michaelcanova: We did not finalize if we wanted it or not in the call, but for now just incase I am removing the up/downvotes from the risk score system.  Thank you again!

src/risk_score/models.py
michaelcanova: Done and updated, thank you again!

src/risk_score/constants.py
yattias: Yeah, makes sense in second thought that this is an optimization we can introduce later on once we encounter a problem that leads to needing to cap a score

yattias (approved):

PR #3458, opened 2026-05-15
Risk Score Services

michaelcanova:
## Overview ##

This PR adds the service files and tests for the new risk score system.


src/risk_score/services.py
```diff
@@ -0,0 +1,84 @@
...
+from risk_score.constants import DEFAULT_SCORE, RESTRICTED_THRESHOLD, TRUSTED_THRESHOLD
+from risk_score.models import RiskScore, RiskScoreEvent
+
+EventType = RiskScoreEvent.EventType
+
+
+class RiskScoreService:
+    """Service for managing user risk score calculations and event recording."""
+
+    def get_score(self, user):
```
yattias, line 14: This name can be misleading to the user. If we are to `get_or_create` then we should name it as such

yattias: I could be wrong here but I actually don't think we should be creating score in a getter now that I think of it. 

Instead, we should create score if it doesn't exist in `record_event`.

In this getter, if risk score doesn't exist we should return an exception, 0 or default value.

However, this really depends on how we are going to use it in the future

michaelcanova: Updated to keep it as a get only, and not a create, and just return a default score if the record does not exist yet.  Thank you for the feedback!

yattias (approved):

PR #3461, opened 2026-05-17
Risk Score Backfill

michaelcanova:
## Overview ##

This PR creates the backfill for the user risk scores, with idempotency to allow it to be run multiple times.


src/risk_score/management/commands/backfill_risk_scores.py
```diff
@@ -0,0 +1,161 @@
...
+
+EventType = RiskScoreEvent.EventType
+
+HISTORICAL_SOURCES = [
+    (Grant, {"status": Grant.OPEN}, EventType.WORK_APPROVED),
+    (Grant, {"status": Grant.DECLINED}, EventType.WORK_DECLINED),
+    (RhCommentModel, {"is_removed": True}, EventType.CONTENT_CENSORED),
+    (ResearchhubPost, {"unified_document__is_removed": True}, EventType.CONTENT_CENSORED),
+    (BountySolution, {"status": BountySolution.Status.AWARDED}, EventType.BOUNTY_AWARDED),
+    (Review, {"is_assessed": True}, EventType.PEER_REVIEW_ASSESSED),
```
yattias, line 47: This is missing tips by the foundation but maybe your code takes it into account already?

src/risk_score/management/commands/backfill_risk_scores.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if already_exists:
+            return
+
+        if not dry_run:
+            self.service.record_event(user, event_type, source=source)
+        self.events_recorded += 1
+
+    def _build_one_time_lookups(self):
+        age_threshold = timezone.now() - timedelta(days=90)
```
yattias: if this is to be used as a bonus we should name it as such

yattias (approved):

src/risk_score/management/__init__.py
gzurowski, line 1: Nit: I don't think this is needed.

src/risk_score/tests/test_backfill_command.py
```diff
@@ -0,0 +1,361 @@
...
+from risk_score.models import RiskScoreEvent
+from risk_score.services import RiskScoreService
+from user.models import UserVerification
+from user.tests.helpers import create_user
+
+EventType = RiskScoreEvent.EventType
+DELTAS = RiskScoreEvent.DELTAS
+
+
+class BackfillCommandMixin:
```
gzurowski, line 29: More of a general comment: For one-off commands, it's probably not meaningful to write elaborate test suites. If it's more than a one-off, the actual business logic should be closer to a service, and the management command just a shell for execution.

gzurowski (approved):

src/risk_score/management/commands/__init__.py
gzurowski, line 1: Nit: Also not needed

src/risk_score/tests/test_backfill_command.py
michaelcanova: Okay thanks, good to know and thank you for the feedback!

PR #3483, opened 2026-05-19
Sending basic data through moderation with grade score

michaelcanova:
## Overview ##

This PR will returns the risk score and a graded score letter for the user moderation tab.  Future PRs will add a full audit of all risk score related actions.

This PR also migrates everything for the risk score from its own app to the user app.  It will be in 2 phases, one for updating the migration, and the second will remove the risk_score app folder completely after the migration updates have been applied.


src/user/serializers.py
```diff
@@ -103,6 +109,15 @@ def get_verification(self, user):
             "status": user_verification.status,
         }
 
+    def get_risk_score(self, user):
+        try:
+            return user.risk_score.score
+        except RiskScore.DoesNotExist:
+            return DEFAULT_SCORE
```
gzurowski: This duplicates the logic that is already present in the risk score service.

gzurowski (commented):
I am not sure if `risk score` deserves to be its own standalone module (Django app). This PR makes it clear by introducing imports that reverse the intended dependency direction. `risk score` currently depends on user, which seems correct (`risk score` -> `user`), because `user` is a lower module. With this PR, `user` now depends back on `risk score` (`user` -> `risk score`).
Since risk score is a concept tightly coupled to the user domain, I don't see much value in having it a standalone module. I would rather integrate risk score in the user app which will keep things cohesive and avoid future circular dependency problems.

src/user/constants/risk_score_constants.py
```diff
@@ -0,0 +1,25 @@
+DEFAULT_SCORE = 100
+TRUSTED_THRESHOLD = 50
+RESTRICTED_THRESHOLD = 150
+
+GRADE_SCALE = [
```
yattias: I'd remove this one for now until we understand scores a bit more. Maybe once we have a good feel for it, we can reintroduce these. Until then, we should send something like `risk_label`:  `TRUSTED` | `RESTRICTED`

yattias (approved):

PR #3485, opened 2026-05-20
Prioritize manually-added experts in search results

koutst:


src/research_ai/serializers.py
```diff
@@ -395,10 +396,21 @@ def get_work(self, obj):
...
+                is_manual=Case(
+                    When(
+                        expert__sources__contains=[{"type": "manual"}],
+                        then=Value(True),
+                    ),
+                    default=Value(False),
+                    output_field=BooleanField(),
+                )
+            )
+            .order_by("-is_manual", "position")
```
koutst: is_manual should probably be a field on expert or search expert but I think we may want to revisit the overall model at some point

nicktytarenko: agree. it should be a seprate field on the Expert 👍🏼

nicktytarenko (approved):
LGTM

src/research_ai/serializers.py
koutst: Ok, I can update. The only issue I saw there was if the expert came up in the future for another search by a different user it could float to the top.

PR #3491, opened 2026-05-20
Remove documents for suspended users

gzurowski:
This change adds signal handlers for user-suspension events to the search app. When a published suspension event is handled, a new task is invoked that updates all search documents (indexes) related to the user.

### DIFF
--- a/src/search/documents/person.py
+++ b/src/search/documents/person.py
@@ -134,6 +134,15 @@ class PersonDocument(BaseDocument):
 
         return suggestions
 
+    @override
+    def should_index_object(self, obj: Author) -> bool:  # type: ignore[override]
+        if obj.user is None:
+            # Leave authors without user accounts in the index:
+            return True
+
+        # only index if the user is not suspended:
+        return not obj.user.is_suspended
+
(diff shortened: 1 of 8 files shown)


src/search/documents/person.py
```diff
@@ -134,6 +134,15 @@ def prepare_suggestion_phrases(self, instance) -> list[dict[str, int]]:
 
         return suggestions
 
+    @override
+    def should_index_object(self, obj: Author) -> bool:  # type: ignore[override]
+        if obj.user is None:
+            # Leave authors without user accounts in the index:
+            return True
```
yattias, line 141: I'm actually not quite sure its beneficial to include authors without user account. from a UX perspective, searching for authors doesn't come up

yattias (approved):

src/search/documents/person.py
gzurowski: This conditional statement preserves the current logic, i.e., authors without a user are indexed (which is the large majority of authors in the database). I didn't want to make unrelated changes to the current logic in this PR.

PR #3493, opened 2026-05-20
Fix burn failures, fix intermitent withdrawal failures

koutst:
BASE withdrawals were occasionally stalling in the mempool because the Etherscan-quoted gas price could be undercut by the time the transaction was submitted. This applies a 20% buffer to the quoted gas price so RSC burn transactions stay competitive through brief fee spikes.

### DIFF
--- a/src/ethereum/lib.py
+++ b/src/ethereum/lib.py
@@ -132,7 +132,8 @@ def execute_erc20_transfer(
     to,
     amount,
     network="ETHEREUM",
-    gas_price=None,
+    max_fee_per_gas=None,
+    max_priority_fee_per_gas=None,
 ):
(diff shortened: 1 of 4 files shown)


src/ethereum/lib.py
```diff
@@ -139,6 +146,8 @@ def execute_erc20_transfer(
         amount (int) - Amount of token to send (in smallest possible
             denomination)
         network (str) - Network to use ("ETHEREUM" or "BASE")
+        gas_price (int) - Optional gas price in wei. If omitted, web3.py's
```
gzurowski: Nit: What is `web3.py` - a file or the `web3` library? Is it possible to add a link to the relevant documentation?

gzurowski (approved):

PR #3513, opened 2026-05-23
Lower fundraise RSC contribution minimum

nicktytarenko:
## What?
- Lower the minimum RSC fundraise contribution from **10 RSC** to **0.01 RSC** so small USD amounts (e.g. $1) are not blocked.


yattias (approved):
Approved but the str to decimal conversion looks weird

PR #3520, opened 2026-05-25
Risk Score Signals

michaelcanova:
## Overview ##

This PR sets up the signals to add records to a user's risk score.  It also removes the old risk_score folder after the migration updates, does some small optimizations for the new signals, and adds tests.


src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+
+_service = RiskScoreService()
+
+
+@lru_cache(maxsize=1)
+def _comment_content_type_id():
+    return ContentType.objects.get_for_model(RhCommentModel).id
+
+
+@receiver(post_save, sender=Grant, dispatch_uid="risk_score_on_grant_status_changed")
```
yattias: For all your signals below, you should include a `transaction.on_commit` to guarantee there is no race condition.

Django fires the `post_save` event before the record is actually saved so there are times where querying the instance may throw an error because it does no exist yet (even though the `instance` variable is in scope!)

yattias: You can look at other signal files such as the ones in personalize app to see how its typically done

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+@receiver(
+    post_save,
+    sender=ResearchhubUnifiedDocument,
+    dispatch_uid="risk_score_on_document_censored",
+)
+def on_document_censored(sender, instance, **kwargs):
+    try:
+        if not instance.is_removed:
+            return
+        author = instance.created_by
```
yattias: author should always exist. If it doesn't, code can flow into exception. We don't need to check `if author is None` I think

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+    post_save,
+    sender=BountySolution,
+    dispatch_uid="risk_score_on_bounty_solution_awarded",
+)
+def on_bounty_solution_awarded(sender, instance, **kwargs):
+    try:
+        if instance.status != BountySolution.Status.AWARDED:
+            return
+
+        if instance.created_by_id:
```
yattias: similar to my comment above. instance should always have a record of who created it. If it doesn't for some reason, it can flow into the exception clause

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+            _service.record_event(
+                instance.created_by, EventType.BOUNTY_AWARDED, source=instance
+            )
+
+        if instance.content_type_id != _comment_content_type_id():
```
yattias, line 146: I'm not sure if we need this

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+    dispatch_uid="risk_score_on_bounty_solution_awarded",
+)
+def on_bounty_solution_awarded(sender, instance, **kwargs):
+    try:
+        if instance.status != BountySolution.Status.AWARDED:
+            return
+
+        if instance.created_by_id:
+            _service.record_event(
+                instance.created_by, EventType.BOUNTY_AWARDED, source=instance
```
yattias: Seems like this signal is creating multiple events. I'd double check if it's the right move

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+        if review.created_by_id:
+            _service.record_event(
+                review.created_by, EventType.PEER_REVIEW_ASSESSED, source=review
+            )
+
+
+@receiver(post_save, sender=Purchase, dispatch_uid="risk_score_on_community_tip")
+def on_community_tip(sender, instance, created, **kwargs):
+    try:
+        if not created:
```
yattias: Same as my comment above about checking for certain conditions that are unlikely to exist

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+    except Exception:
+        logger.exception("Risk score signal failed for Purchase %s", instance.pk)
+
+
+@receiver(
+    post_save,
+    sender=SocialAccount,
+    dispatch_uid="risk_score_on_social_account_created",
+)
+def on_social_account_created(sender, instance, created, **kwargs):
```
yattias, line 183: I'm not sure honestly if we need an event for an orcid verified edu email when we have EDU_EMAIL_SIGNUP

Maybe just a generic event for `EDU_EMAIL` that can be triggered by both signals

src/user/signals/risk_score_signals.py
```diff
@@ -0,0 +1,223 @@
...
+@receiver(
+    post_save,
+    sender=UserVerification,
+    dispatch_uid="risk_score_on_persona_verified",
+)
+def on_persona_verified(sender, instance, **kwargs):
+    try:
+        if instance.status != UserVerification.Status.APPROVED:
+            return
+        # PERSONA_VERIFIED_NON_WHITELISTED is reserved for when UserVerification
```
yattias, line 208: Just wanted to flag this to you as a reminder we will need to update persona flow to add country. You can ask @gzurowski to help with this so you can focus on finishing this up

src/user/tasks/risk_score_tasks.py
```diff
@@ -0,0 +1,31 @@
...
+from researchhub.celery import app
+from user.models import User
+from user.related_models.risk_score_model import RiskScoreEvent
+from user.services.risk_score_service import RiskScoreService
+
+EventType = RiskScoreEvent.EventType
+
+
+@app.task
+def apply_account_age_bonus_task():
```
yattias, line 14: This is OKAY but let's make sure it doesn't run more than once a day. You will need to configure it in `celery.py`

src/user/signals/risk_score_signals.py
michaelcanova: This is actually required. Since reviews only exists on comments, this just makes sure we don't try to grab `item.reviews` on a bounty solution that's a paper or post and crash. We make the same check already in `review_signals.py`.

src/user/signals/risk_score_signals.py
michaelcanova: I consolidated this down just for EDU_EMAIL, but both receivers still need to exist, as they happen at different times (one when the user signs up, the other when they connect ORCID).  However, it will now only apply as one risk score modifier, rather than 2.

src/user/tasks/risk_score_tasks.py
michaelcanova: Undertsood and celery configuration was added.  thank you!

src/user/tests/test_risk_score_tasks.py
```diff
@@ -0,0 +1,77 @@
...
+        apply_account_age_bonus_task()
+
+        # Assert
+        self.assertFalse(
+            RiskScoreEvent.objects.filter(
+                event_type=EventType.ACCOUNT_AGE_BONUS
+            ).exists()
+        )
+
+    def test_skips_inactive_users(self):
```
yattias: Let's be more specific with test name: `test_no_age_bonus_for_inactive_users

yattias (approved):

PR #3529, opened 2026-05-29
Risk Score Audit API

michaelcanova:
## Overview ##

This PR creates the API endpoints for the user moderation page to show risk score, insights, and risk score records with appropriate details.


src/user/views/moderator_view.py
yattias, line 1: Can you add a test to ensure only moderators have access to this endpoint?

michaelcanova: Added the tests.  Thank you again!

src/user/services/risk_score_insights_service.py
```diff
@@ -0,0 +1,283 @@
...
+            buckets[key] = {
+                "count": 0,
+                "total_delta": 0,
+                "min_delta": agg["min_delta"],
+                "max_delta": agg["max_delta"],
+            }
+        bucket = buckets[key]
+        bucket["count"] += agg["count"]
+        bucket["total_delta"] += agg["total_delta"]
+        bucket["min_delta"] = min(bucket["min_delta"], agg["min_delta"])
```
yattias, line 249: seems like min_delta and max_delta are used to determine the sentiment of the bucket. 
1. Do we need the sentiment per bucket? what is it used for? Is it to render if the up or down arrow on the frontend?
2. If so, seems like we need `has_negative` and `has_positive` out of it instead the values. If so, maybe we can get rid of the `_sentiment` function and just add these keys directly to the bucket

src/user/services/risk_score_insights_service.py
```diff
@@ -0,0 +1,283 @@
...
+from researchhub_document.related_models.researchhub_post_model import ResearchhubPost
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+from review.models.review_model import Review
+from user.related_models.risk_score_model import RiskScoreEvent
+
+logger = logging.getLogger(__name__)
+
+SNIPPET_LENGTH = 200
```
yattias: I wonder if we need snippet. Typicaly truncation of text happens on the FE unless we are talking about massive data fields

src/user/services/risk_score_insights_service.py
yattias, line 1: In general the code in this file is good but wondering how much of it can be controlled by the FE

src/user/services/risk_score_insights_service.py
```diff
@@ -0,0 +1,283 @@
...
+
+    return {
+        event.id: _build_detail(
+            sources.get((event.source_content_type_id, event.source_content_id))
+        )
+        for event in events
+    }
+
+
+def _sentiment(min_delta, max_delta):
```
yattias: I wonder if we need negative, positive here. the frontend can determine that based on score, no? maybe there is a reason

src/user/services/risk_score_insights_service.py
michaelcanova: I simplified it further just returning the min and max deltas and having the front end determine from there.  thank you!

yattias (approved):

src/user/services/risk_score_insights_service.py
michaelcanova: moved truncation logic to the FE, thank you again!

PR #3547, opened 2026-06-01
Reduce expert finder batch size and report failures to Sentry

nicktytarenko:
## What?
- Lower per-round expert discovery batch size and allow more fill rounds to reduce truncated LLM JSON (e.g. `JSON parse failed` on large `expert_count` searches).
- Report all expert finder soft failures to Sentry 

## How?
- Set `EXPERT_FILL_BATCH_MAX` to `30` (from `50`) and `EXPERT_FILL_MAX_ROUNDS` to `8` (from `6`) so each OpenAI call requests fewer experts per round while still filling up to the target via multiple rounds.
- Call `sentry.log_error` inside `fail_return` on every failure


src/research_ai/services/expert_finder_service.py
```diff
@@ -366,7 +367,22 @@ def fail_return(
...
+            }
+            if sentry_extra:
+                payload.update(sentry_extra)
+            sentry.log_error(
+                sentry_exception
+                if sentry_exception is not None
+                else RuntimeError(current_step),
+                message=f"Expert finder failed: {current_step}",
+                json_data=payload,
+            )
```
gzurowski: Please use `logger.error` which will automatically report into Sentry as well.

PR #3588, opened 2026-06-05
Adding Status to Works

michaelcanova:
## Overview ##

This PR adds the status column to works, both for posts/preregistrations and for papers, to set up for risk score enforcement.  This PR also includes the Mixin that will be utilized for visibility and moderation.


src/paper/migrations/0174_paper_status_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+        migrations.AddField(
+            model_name="paper",
+            name="reviewed_date",
+            field=models.DateTimeField(
+                blank=True,
+                help_text="When the work was approved or declined",
+                null=True,
+            ),
+        ),
+        migrations.AddIndex(
```
koutst: Adding this index to a table to size of the paper table can potentially locking of the table and our site to go down. Two suggestions, 
1. Use AddIndexConcurrently instead of AddIndex
2. Create a partial index since we will only really use that index to find the subset of papers pending or potentially declined, and a full index would be quite big

Ex. 
```
  AddIndexConcurrently(
      model_name="paper",
      index=models.Index(
          fields=["status"],
          name="paper_status_pending_idx",
          condition=Q(status__in=["PENDING", "DECLINED"]),
      ),
  ),
```

src/utils/models.py
```diff
@@ -88,6 +90,47 @@ def delete(self, soft=True, *args, **kwargs):
             return super().delete(*args, **kwargs)
 
 
+class ModeratedDocumentMixin(models.Model):
```
koutst, line 93: Instead of adding this mixin to both papers and posts, would it make more sense to have it on the unified document level?

src/utils/models.py
```diff
@@ -88,6 +90,47 @@ def delete(self, soft=True, *args, **kwargs):
...
+    DECLINED = "DECLINED"
+    STATUS_CHOICES = (
+        (PENDING, "Pending"),
+        (APPROVED, "Approved"),
+        (DECLINED, "Declined"),
+    )
+
+    status = models.CharField(
+        choices=STATUS_CHOICES,
+        default=APPROVED,
```
koutst, line 112: This should probably be updated after initial release so we don't accidentally approve papers and posts if we forget to set the field

michaelcanova: Makes perfect sense to me and I agree!  Thank you!

src/paper/migrations/0174_paper_status_and_more.py
michaelcanova: Updated and done, thank you!

src/utils/models.py
michaelcanova: Updated and done, thank you!

koutst (approved):

PR #3589, opened 2026-06-05
Work Moderation APIs

michaelcanova:
## Overview ## 

This PR accomplishes several tasks, including:

- Implements moderation queue APIs for all work types
- Sending notifications for works that are approved/declined
- Defers publishing of feed items until after approval
- Added proper type hinting to risk score functions that were missing it.
- Optimized functions for DRYness
- Added testing


src/feed/signals/document_signals.py
```diff
@@ -71,6 +75,11 @@ def _create_document_feed_entries(instance, pk_set):
         for document_id in pk_set:
             unified_document = hub.related_documents.get(id=document_id)
             item = unified_document.get_document()
+
+            # Works awaiting moderation publish on approval, not when hubs change.
+            if not unified_document.is_approved:
```
yattias: I have a suggestion, and it goes also for the change above:

Rather than having the `is_approved` check scattered across different conditionals, I'd suggest you create a helper function (could even be scoped within this signal) that is called at the top. Could be something like `is_approved_by_moderation`. Within this function, it is okay to put all of the conditionals you need in order to check whether it is approved even if you may be repeating yourself - I think this one of the few cases worth repeating yourself for the sake of consolidation and readability. That way if an AI or human looking at the code they can mentally "move on" knowing that moderation check is done and they can focus on the rest of the code.

Another important consideration we should have @koutst or @gzurowski chime in is on whether this conditional can prevent preprints automatically being created in our systems (e.g. biorxiv, arxiv, ..)

src/feed/views/feed_view.py
```diff
@@ -48,6 +72,76 @@ def list(self, request, *args, **kwargs):
 
         return self._get_non_personalized_feed_response(request, feed_view)
 
+    @action(detail=False, methods=["get"], permission_classes=[IsModerator])
+    def pending_moderation(self, request: Request) -> Response:
```
yattias, line 76: I mentioned this in the other PR but just to ensure this doesn't get missed, we should move moderator related endpoints to its own view

src/feed/views/feed_view_mixin.py
```diff
@@ -62,6 +65,27 @@ def _get_content_type(self, model_class):
             )
         return self._content_types[model_name]
 
+    @staticmethod
+    def build_unsaved_feed_entry(
```
yattias, line 69: Nice. I like the refactor.

src/feed/tasks.py
```diff
@@ -56,6 +57,16 @@ def create_feed_entry(
     if unified_document is not None and not unified_document.is_public:
         return None
 
+    # Defense-in-depth: never publish a work still awaiting moderation. The
+    # publishing signals already defer these, but the task can be called directly.
+    if (
```
yattias, line 62: This is a good check for correctness but similar to my comment at the very top, we need to ensure the new condition `not unified_document.is_approved` doesn't break the preprint pipeline. CC @gzurowski @koutst

src/notification/migrations/0024_alter_notification_notification_type.py
yattias, line 1: This is good but just a comment for future, the notification related changes could go out separately in a different PR. No need to change anything but since we are trying to improve on iterative dev, we could get this part out before the rest of the changes since nothing would trigger them

src/paper/views/paper_views.py
```diff
@@ -35,7 +34,15 @@
     PaperSerializer,
     PaperSubmissionSerializer,
 )
-from user.permissions import IsVerifiedUser
+from paper.utils import get_cache_key
+from reputation.related_models.paper_reward import (
```
yattias: I think a bunch of these imports were added but doesn't seem like they were used

michaelcanova: this was from a merge conflict, it has now been removed.  thank you!

src/feed/signals/document_signals.py
michaelcanova: I split the section up to reusable components, thank you!

src/feed/tasks.py
michaelcanova: It is actually fine, because the pipeline right now doesn't pass a status in, and in the mixin if there is no status passed in, it defaults to APPROVED.  so this will keep functionality working as intended.

src/feed/views/feed_view.py
michaelcanova: It is done 100% in the later branch, and will be applied when we merge, since all will be merged at the same time.  thank you!

yattias (approved):

PR #3591, opened 2026-06-06
Add dual RSC/USD amounts to FundingActivity

nicktytarenko:
## What?
- Add `usd_cents` to `FundingActivity` and `FundingActivityRecipient`, plus `USD_FUNDRAISE_PAYOUT` source type
- Populate both RSC and USD amounts when new funding activities are created

## How?
- Migration adds `usd_cents` column
- New `funding_activity_amounts` helpers resolve rates and convert between RSC and USD cents
- Update `funding_activity_service` create paths to set both legs at write time (Purchase rate, then historical fallback)


src/user/related_models/funding_activity_model.py
```diff
@@ -109,18 +115,22 @@ class FundingActivityRecipient(DefaultModel):
...
-        help_text="The user who received the funding",
+        db_comment="The user who received the funding",
     )
     amount = models.DecimalField(
         max_digits=19,
         decimal_places=8,
-        help_text="RSC amount received by this recipient",
+        db_comment="RSC amount received by this recipient",
+    )
+    usd_cents = models.IntegerField(
```
gzurowski:
```suggestion
    amount_usd_cents = models.IntegerField(
```
Should be align the names of both amount fields?

src/user/services/funding_activity_amounts.py
```diff
@@ -0,0 +1,59 @@
+from decimal import Decimal
+
+from purchase.related_models.rsc_exchange_rate_model import RscExchangeRate
+
+
+class FundingActivityAmountsService:
```
gzurowski: Nit: Do we really need a dedicated service for amounts? Aren't these just helper methods for the existing funding activity service?

gzurowski (approved):

PR #3592, opened 2026-06-06
Add USD_FUNDRAISE_PAYOUT funding activity create path

nicktytarenko:
## What?
- Wire `USD_FUNDRAISE_PAYOUT` into the FundingActivity ledger for non-refunded USD fundraise contributions when a fundraise escrow completes
- Store native `usd_cents` plus calculated RSC at write time using the historical rate at contribution time

## How?
- Add `get_usd_fundraise_payouts()` and `_create_usd_fundraise_payout_activity()` to `FundingActivityService`
- On escrow `PAID` + `FUNDRAISE`, schedule `USD_FUNDRAISE_PAYOUT` tasks for each non-refunded `UsdFundraiseContribution`


src/user/services/funding_activity_service.py
```diff
@@ -270,6 +292,52 @@ def _create_fundraise_payout_activity(cls, purchase) -> Optional[FundingActivity
...
+            )
+            .select_related("escrow", "unified_document")
+            .first()
+        )
+        if not fundraise or not fundraise.escrow:
+            return None
+        try:
+            recipient_user = fundraise.get_recipient()
+        except Exception:
+            return None
```
gzurowski, line 357: Is it OK that these errors go totally unnoticed? If yes, there should be a comment or explanation in the function docs.

src/user/services/funding_activity_service.py
```diff
@@ -270,6 +292,52 @@ def _create_fundraise_payout_activity(cls, purchase) -> Optional[FundingActivity
...
+        )
+        rate = FundingActivityAmountsService.get_historical_rsc_usd_rate(
+            contribution.created_date
+        )
+        FundingActivityAmountsService.populate_usd_native_dual_amounts_on_recipients(
+            activity, [recipient], native_usd_cents, rate
+        )
+        activity.save()
+        recipient.activity = activity
+        recipient.save()
```
gzurowski: Does this need to be wrapped in a transaction?

gzurowski (approved):

src/user/services/funding_activity_service.py
nicktytarenko: Those saves are already inside a transaction _(in the parrent method)_

PR #3613, opened 2026-06-10
Add viewer-segment-aware caching to grant and funding feeds

nicktytarenko:
## What?
- Introduce viewer-segment-aware cache keys so public discovery traffic stays on a shared cache;
- Users with private visibility get per-user cache entries;
- Moderators/hub editors always receive fresh responses.

## WHy?
- Prepares both feeds for serving private content correctly later without one global cached payload for everyone.

## How?
- Add `get_feed_cache_segment()` to resolve the viewer segment:

  - `:public` => anonymous users, or logged-in users with no private post visibility
  - `:viewer-{user_id}` => users who can see private posts via `ResearchhubPost.visible_to()`
  - **No cache** =>  moderators and hub editors

### DIFF
--- a/src/feed/cache_segment.py
+++ b/src/feed/cache_segment.py
@@ -1,0 +1,34 @@
+from rest_framework.request import Request
+
+from researchhub_document.related_models.researchhub_post_model import ResearchhubPost
+
+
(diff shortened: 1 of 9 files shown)


koutst (approved):
LGTM we should just keep an eye out for performance issues for all theses cases now

PR #3614, opened 2026-06-10
Risk Score Enforcement

michaelcanova:
## Overview ## 

This PR accomplishes implements enforcement and signals for risk score changes across all work types.  It makes decisions on whether a work will be pending for moderation, or auto accepted, based on user risk scores, and updates opensearch and feeds based on the work status.  It also prevents items from being viewable by users other than the creator, moderators, and hub editors until until moderation approves the work.  


src/feed/tests/views/test_funding_feed_view.py
```diff
@@ -189,6 +189,30 @@ def test_list_funding_feed(self):
         self.assertNotIn(self.non_preregistration_post.id, post_ids)
         self.assertNotIn(self.removed_post.id, post_ids)
 
+    def test_pending_preregistration_excluded(self):
```
yattias: Let's be more verbose with naming so that future tests AI helps generate are verbose and easy to read. `test_pending_preregistration_excluded_from_funding_feed` is good

src/feed/tests/views/test_journal_feed_view.py
```diff
@@ -148,6 +148,36 @@ def test_list_journal_feed(self):
         self.assertNotIn("Non-Journal Paper", titles)
         self.assertNotIn("Removed Journal Paper", titles)
 
+    def test_pending_paper_excluded(self):
```
yattias: `test_pending_paper_excluded_from_journal_feed`

src/feed/tests/views/test_pending_moderation.py
```diff
@@ -0,0 +1,152 @@
...
+from researchhub_document.related_models.researchhub_unified_document_model import (
+    ResearchhubUnifiedDocument,
+)
+from user.constants.risk_score_constants import DEFAULT_SCORE
+from user.related_models.risk_score_model import RiskScore
+from user.tests.helpers import create_random_default_user
+from utils.test_helpers import AWSMockTestCase
+
+
+class PendingModerationRiskScoreTests(AWSMockTestCase):
```
yattias: In all your tests, I don't think we need AWSMockTestCase. That's used for specific reasons. I think `APITestCase` should be used instead. I think AI keeps trying to propegate `AWSMockTestCase` and we should prevent it from doing so by using the right one.

The ask here is to update all your tests to use `APITestCase` instead

src/feed/views/feed_view.py
```diff
@@ -28,8 +28,12 @@
     ResearchhubUnifiedDocument,
 )
 from user.permissions import IsModerator
+from user.related_models.risk_score_model import RiskScore
 from utils.throttles import FeedRecommendationRefreshThrottle
 
+if TYPE_CHECKING:
```
yattias: ???

I'd double check this is needed? Not sure the purpose

src/feed/views/feed_view.py
```diff
@@ -89,27 +93,74 @@ def pending_moderation(self, request: Request) -> Response:
...
+        return Response(
+            {
+                "funding_opportunities": Grant.objects.filter(
+                    status=Grant.PENDING
+                ).count(),
+                "proposals": pending_posts.filter(
+                    document_type=PREREGISTRATION
+                ).count(),
+                "posts": pending_posts.filter(document_type=DISCUSSION).count(),
+                "journal_entries": Paper.objects.filter(
```
yattias: Even though status is indexed, it is not guaranteed that querying the table will be fast. 
Indexes work really well for unique entities like IDs but less so for enums like statuses because there is no variety to divide the dataset on - only two or three things. 

For reference, we also have a `is_removed` index on this table but running `Paper.objects.filter(is_removed=False).count()` (without your status filter since it doesn't exist on prod) took about 3 minutes.

If I took a calculated bet I'd say this would lock our DB within 10 minutes of being released.

src/feed/views/feed_view.py
```diff
@@ -89,27 +93,74 @@ def pending_moderation(self, request: Request) -> Response:
...
+        )
         return self.get_paginated_response(serializer.data)
 
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="pending_moderation/counts",
+        permission_classes=[IsModerator],
+    )
+    def pending_moderation_counts(self, request: Request) -> Response:
```
yattias: I think we should extract all of the moderator related endpoints out of the `feed_view` as to not complicate a public feed with moderator specific one. 

You can create a new feed view called `moderator_feed_view` and move all moderator functionality in there. 

Permission classes could then be set on the class level rather on the endpoint level

src/feed/views/feed_view.py
```diff
@@ -89,27 +93,74 @@ def pending_moderation(self, request: Request) -> Response:
...
+            for item, author in zip(page, authors)
         ]
-        serializer = self.get_serializer(feed_entries, many=True)
+        serializer = self.get_serializer(
+            feed_entries,
+            many=True,
+            context={
+                **self.get_serializer_context(),
+                "include_risk_score": True,
+                "risk_score_by_user_id": self._risk_score_by_user_id(authors),
```
yattias: See my comment below regarding creating a moderator feed. 

Similarly, I think we should created a `ModeratorFeedEntrySerializer` that is super simple and inherits from `FeedEntrySerializer`. That way you will not have to pass `include_risk_score` as context and further more, `risk_score_by_user_id` functionality can live on the serializer itself since it is the responsibility of the serializer

src/feed/views/funding_feed_view.py
```diff
@@ -123,8 +123,10 @@ def get_queryset(self):
             queryset = queryset.filter(id__in=visible_ids)
         else:
             # The public discovery feed stays user-agnostic so it can be cached
-            # for everyone; never expose private work here.
-            queryset = queryset.filter(unified_document__is_public=True)
+            # for everyone; never expose private or unmoderated work here.
+            # visible_to(None) is the canonical public predicate (is_public +
+            # approved), shared with the per-viewer branch above.
+            queryset = queryset.visible_to(None)
```
yattias: I like the `visible_to` conceptually on the queryset, but passing `None` as a filtering param feels weird. Don't have a great suggestion but wondering if there is something more clear we can do here

src/researchhub/serializers/serializers.py
```diff
@@ -1,6 +1,22 @@
 import rest_framework.serializers as serializers
 
 
+class ModeratedDocumentStatusSerializerMixin(serializers.Serializer):
```
yattias, line 4: Nice  way to keep things DRY

src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -31,20 +31,36 @@ class ResearchhubPostQuerySet(models.QuerySet):
     def visible_to(self, user):
```
yattias: We should add type hints, seeing what function returns is really helpful before reading code

src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -31,20 +31,36 @@ class ResearchhubPostQuerySet(models.QuerySet):
...
+        A post that has NOT cleared moderation is visible only to its author and
+        to site moderators / hub editors (who need to see it to moderate it).
+        The broader "elevated" viewers -- the creator of any grant the post
+        applied to (via GrantApplication) and users with a non-revoked
+        Permission on the post's unified document (e.g. invited experts) -- may
+        only see the post once it has cleared moderation. A NO_ACCESS Permission
+        row revokes access even when other (e.g. stale VIEWER) rows exist for the
+        same user/document.
+
+        GRANT posts stay APPROVED regardless of the grant's moderation state, so
```
yattias: This is a big comment and it is helpful but too long to read. Can you make it more concise? I'd add more plainly that grant is basically excluded from unified doc status since its got its own

src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -31,20 +31,36 @@ class ResearchhubPostQuerySet(models.QuerySet):
...
         """
-        public = Q(unified_document__is_public=True)
+        pending_grant = Grant.objects.filter(
+            unified_document_id=OuterRef("unified_document_id"),
+            status__in=Grant.PENDING_MODERATION_STATUSES,
+        )
+        cleared = Q(
+            unified_document__status=ResearchhubUnifiedDocument.APPROVED
+        ) & ~Exists(pending_grant)
+        public = Q(unified_document__is_public=True) & cleared
```
yattias: let's rename `is_public` to explicitly convey that this is a boolean

src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -31,20 +31,36 @@ class ResearchhubPostQuerySet(models.QuerySet):
...
+        GRANT posts stay APPROVED regardless of the grant's moderation state, so
+        they gate on ``Grant.status`` instead: a pending or declined grant is
+        treated as not yet cleared.
         """
-        public = Q(unified_document__is_public=True)
+        pending_grant = Grant.objects.filter(
+            unified_document_id=OuterRef("unified_document_id"),
+            status__in=Grant.PENDING_MODERATION_STATUSES,
+        )
+        cleared = Q(
```
yattias: To be more verbose, let's change to `moderation_approved`. 
`cleared` ain't bad but the explicit language helps a lot when reading

src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -56,12 +72,12 @@ def visible_to(self, user):
         allowed = user_perms.exclude(access_type=NO_ACCESS)
         revoked = user_perms.filter(access_type=NO_ACCESS)
 
-        return self.filter(
-            public
-            | Q(created_by=user)
-            | Q(grant_applications__grant__created_by=user)
+        elevated = cleared & (
```
yattias: something about `elevated` doesn't feel right. Can we try something else such as explicitly filtering for moderation_approved OR creator_of_document.

src/paper/related_models/paper_model.py
```diff
@@ -36,12 +37,37 @@
 logger = logging.getLogger(__name__)
 
 
+class PaperQuerySet(models.QuerySet):
+    def visible_to(self, user):
```
yattias: Let's have type hint in here

src/purchase/services/grant_service.py
```diff
@@ -73,6 +75,25 @@ def _publish_to_feed(self, post):
         except Exception:
             logger.exception("Failed to create feed entry for post %s", post.id)
 
+    def _reindex_post(self, post):
```
yattias, line 78: The opensearch part was not tested by me yet. This looks okay for now if it works. I would like to get @gzurowski eyes on it but it should not be merge-blocking

src/paper/views/paper_views.py
```diff
@@ -251,6 +263,11 @@ def create_researchhub_paper(self, request):
                     )
                     raise
 
+                # The Paper post_save signal creates the unified document, which
```
yattias, line 255: @koutst this makes sense here logically to properly set the status as the document is created. However, wanted to flag this in case you have a better idea which will not couple this endpoint further

src/user/tests/test_risk_score_enforcement.py
```diff
@@ -0,0 +1,100 @@
...
+from researchhub_document.views.researchhub_post_views import (
+    MIN_POST_BODY_LENGTH,
+    MIN_POST_TITLE_LENGTH,
+)
+from user.related_models.risk_score_model import RiskScore
+from user.services.risk_score_service import RiskScoreService
+from user.tests.helpers import create_random_default_user, make_user_verified
+from utils.models import ModeratedDocumentMixin
+
+NORMAL = 100
```
yattias: Even better if we can use existing constants

src/paper/views/paper_views.py
koutst: This  seems fine to me. Alternatively we could have a unified document signal that handles risk scores on the related document (post or paper)

yattias: Okay lets leave it then

yattias (approved):

PR #3642, opened 2026-06-12
Add OA full-text fields and filter to OpenAlex utils

koutst:


src/utils/tests/openalex_helpers.py
```diff
@@ -0,0 +1,43 @@
+"""Shared factories for building OpenAlex API payloads in tests."""
+
+
+def oa_author_record(**overrides):
```
gzurowski:
```suggestion
def create_oa_author_record(**overrides):
```

src/utils/tests/openalex_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+        "works_count": 40,
+        "cited_by_count": 900,
+        "affiliations": [{"institution": {"display_name": "Stanford University"}}],
+        "topics": [{"display_name": "Genomics"}, {"display_name": "Bioinformatics"}],
+    }
+    record.update(overrides)
+    return record
+
+
+def oa_work(
```
gzurowski:
```suggestion
def create_oa_work(
```

gzurowski (approved):

PR #3660, opened 2026-06-15
Add llm search for disambiguation

koutst:
This PR adds LLM-powered author disambiguation to the researcher-profile pipeline. It introduces an escalation-ladder resolver that maps an expert to an OpenAlex author through progressively less-certain rungs — a cited ORCID source link, an institution-scoped name+affiliation match, and finally an OpenAI web-search disambiguator that picks a candidate, reports an identifier it found online, or abstains. Supporting changes refactor shared LLM-JSON parsing and ORCID normalization into reusable helpers, harden name matching, and add substantial test coverage across the new resolver, disambiguator, and LLM service.


src/research_ai/services/openai_llm_service.py
```diff
@@ -0,0 +1,118 @@
+import logging
+
+from django.conf import settings
+from openai import OpenAI
+
+from utils import sentry
```
gzurowski, line 6: We are in the process of deprecating the Sentry utilities. The logging that is already present (error/exception level) already takes care of pushing into Sentry.

gzurowski (approved):

PR #3670, opened 2026-06-18
Auto-exclude experts from prior searches on the same document

nicktytarenko:
## What?
- Removes the client-supplied `excluded_search_ids` parameter from the expert finder create API and database model.
- Automatically excludes experts already linked to any prior expert search on the same unified document when running a new search.

## How?
- Replaces `_names_and_emails_from_excluded_searches` with `_names_and_emails_from_prior_document_searches`, which queries all `SearchExpert` rows for the current search’s `unified_document_id` (excluding the current search ID).

## Breaking changes
- List and detail responses no longer include `excluded_search_ids`. https://github.com/ResearchHub/web/pull/911


src/research_ai/tests/test_tasks.py
```diff
@@ -58,13 +58,10 @@ def setUp(self):
...
             query="Task query",
             status=ExpertSearch.Status.PENDING,
-            excluded_search_ids=[42],
         )
 
     @patch("research_ai.tasks.expert_finder_service_mod.run_expert_finder_search")
-    def test_task_merges_excluded_search_ids_from_model_when_kwargs_omitted(
-        self, mock_run
-    ):
+    def test_task_does_not_pass_excluded_search_ids(self, mock_run):
```
gzurowski: If `excluded_search_ids` is no longer part of the contract, is it worth testing for it?

gzurowski (approved):

src/research_ai/tests/test_tasks.py
nicktytarenko: yes, removed it 🧹

PR #3674, opened 2026-06-22
Adding Research Journey Model & Migration

michaelcanova:
## Overview ##

This PR adds the new research journey model and migrations, which will be the one to many relationship to connect the grant -> preregistration -> registered report flow.


src/researchhub_document/related_models/research_journey_model.py
```diff
@@ -0,0 +1,60 @@
...
+class ResearchJourney(DefaultModel):
+    created_by = models.ForeignKey(
+        "user.User",
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="created_research_journeys",
+        help_text="User who created the source preregistration.",
+    )
+    grant = models.ForeignKey(
```
koutst: Should this just be the post of the grant?

src/researchhub_document/related_models/research_journey_model.py
```diff
@@ -0,0 +1,60 @@
+from django.db import models
+
+from utils.models import DefaultModel
+
+
+class ResearchJourney(DefaultModel):
+    created_by = models.ForeignKey(
```
koutst: Why do we need created by here? Isn't this essentially a join table?

src/researchhub_document/related_models/research_journey_model.py
michaelcanova: good callout, that was a misstep using the purchase table instead of the post.  thank you!

src/researchhub_document/related_models/research_journey_model.py
michaelcanova: good point, we really don't need it at all here.  Removed and updated, thank you!

PR #3676, opened 2026-06-22
Adding Research Journey Service and tests

michaelcanova:
## Overview ##

This PR adds a basic research journey service for stage management and connecting the different portions of the grant -> proposal -> registered report stages.  It also adds testing for these services.


src/researchhub_document/tests/test_journey_service.py
```diff
@@ -0,0 +1,237 @@
...
+            document_type=GRANT,
+        )
+        return Grant.objects.create(
+            created_by=self.user,
+            unified_document=unified_document,
+            amount=Decimal("1000.00"),
+            currency="USD",
+            organization="Research Grant",
+            description="Funding for the proposal.",
+        )
```
gzurowski: Nit: Methods should be verbs, not nouns. E.g., `_create_journey` or `_create_post` would also better reveal the method's intent.

src/researchhub_document/services/journey_service.py
```diff
@@ -0,0 +1,190 @@
...
+)
+
+
+@dataclass(frozen=True)
+class JourneyStage:
+    stage: str
+    item: object
+
+
+class JourneyService:
```
gzurowski, line 27: It would be good to have some brief class-level and method-level documentation.

src/researchhub_document/services/journey_service.py
```diff
@@ -0,0 +1,190 @@
...
+        journey_model=None,
+        post_model=None,
+        grant_application_model=None,
+    ):
+        self.journey_model = journey_model or ResearchJourney
+        self.post_model = post_model or ResearchhubPost
+        self.grant_application_model = grant_application_model or GrantApplication
+
+    @transaction.atomic
+    def get_or_create_for_preregistration(self, post):
```
gzurowski: Adding type annotations for arguments and the return value should be helpful.

src/researchhub_document/tests/test_journey_service.py
michaelcanova: Updated and fixed, thank you!

src/researchhub_document/services/journey_service.py
michaelcanova: Updated and fixed, thank you!

src/researchhub_document/services/journey_service.py
michaelcanova: Good callout, I added docstrings to all functions!  Thank you!

PR #3680, opened 2026-06-23
Create Research Journey on Preregistration Approval

michaelcanova:
## Overview ##

This PR creates the research journey entry when a preregistration is approved, whether through moderaiton or auto approval through the risk score.


src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -337,6 +338,8 @@ def create_researchhub_post(self, request):
                     )
                     GrantCacheMixin.invalidate_grant_feed_cache()
 
+                JourneyService().ensure_approved_preregistration_has_journey(rh_post)
```
koutst, line 341: Should we move this up to where we set work_status?

koutst (approved):

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: While i do agree with keeping it close to the work status, the write needs to stay after GrantApplication creation so the journey captures grant_post

PR #3681, opened 2026-06-23
Completed Preregistrations can show in Journal

michaelcanova:
## Overview ##

This PR sets it where a completed/funded preregistration that has a journey id will be able to be shown in the journal by flipping the `is_in_journal` to true.


src/researchhub_document/services/journey_service.py
```diff
@@ -77,6 +78,34 @@ def ensure_approved_preregistration_has_journey(
...
+        if fundraise.status != Fundraise.COMPLETED:
+            return None
+
+        proposal = self._get_preregistration_for_fundraise(fundraise)
+        if proposal is None:
+            return None
+
+        journey = self.ensure_approved_preregistration_has_journey(proposal)
+        if journey is None:
+            return None
```
koutst, line 116: These would all end in silent failures. If these fail it probably indicates a bug, right? Should we log at least?

koutst (approved):

src/researchhub_document/services/journey_service.py
michaelcanova: Good callout!  I added loggers for 2 of the 3, since those would actually be errors, but the fundraise completed check is fine as is since its expected.  Thank you!

PR #3684, opened 2026-06-23
Create Registered Report

michaelcanova:
## Overview ##

This PR adds the endpoint allowing the creation of registered reports from the notebook.


koutst (approved):

src/researchhub_document/services/journal_entry_service.py
```diff
@@ -0,0 +1,316 @@
...
+        self.fundraise_model = fundraise_model or Fundraise
+        self.journey_service = journey_service or JourneyService()
+        self.note_model = note_model or Note
+        self.note_content_model = note_content_model or NoteContent
+        self.permission_model = permission_model or Permission
+        self.post_model = post_model or ResearchhubPost
+        self.unified_document_model = (
+            unified_document_model or ResearchhubUnifiedDocument
+        )
+        self.user_model = user_model or User
```
gzurowski: The init arguments for this service are overkill. Service dependencies should be injected for improved testability, so the `JourneyService` argument seems fine. The actual models, the application's business model that the internal logic operates on, must not be injected.

gzurowski (changes requested):

src/researchhub_document/services/journal_entry_service.py
michaelcanova: Good callout!  I tried to simplify it, but you are correct, it is better practice to do it this way, so I updated it to the specifications you requested.  Thank you again!

PR #3685, opened 2026-06-23
Registered Report Results

michaelcanova:
## Overview ##

This PR allows the addition of results to a registered report utilizing the same functionality and concepts as an author update on a proposal.


src/researchhub_comment/views/rh_comment_view.py
```diff
@@ -388,6 +402,16 @@ def _create_rh_comment(self, request, *args, **kwargs):
             ).data
             return Response(serializer_data, status=200)
 
+    def validate_registered_report_results_target(self, target: object) -> None:
```
gzurowski: Nit: Can this be a private method?
```suggestion
    def _validate_registered_report_results_target(self, target: object) -> None:
```

gzurowski (approved):

src/researchhub_comment/views/rh_comment_view.py
```diff
@@ -349,6 +354,15 @@ def _create_rh_comment(self, request, *args, **kwargs):
             if not getattr(target, "created_by", None) or target.created_by != user:
                 raise PermissionDenied("Only the author can add author updates.")
 
+        if data.get("thread_reference") == REGISTERED_REPORT_RESULTS_REFERENCE:
```
yattias, line 357: Seems like we are treating results as author updates. Can you explain how this would work exactly from a user perspective? What would be the UX that would create these results?

PR #3694, opened 2026-06-25
Default the profile agent to Opus 4.8 and ground works from OpenAlex truth

koutst:
- Can be adjusted later to reduce cost

### DIFF
--- a/src/research_ai/services/researcher_profile/agent.py
+++ b/src/research_ai/services/researcher_profile/agent.py
@@ -50,9 +50,11 @@ How to work:
   ones where this author is first/last author and that are recent and relevant.
   Only keep works that have a pdf_url (readable full text).
 
-Grounding rule: every work you submit MUST be copied verbatim from a
-get_author_works result -- same title, source_url, and pdf_url. Never invent or
-edit a URL. Finish by calling submit_profile exactly once.
+Grounding rule: every work you submit MUST come from a get_author_works result.
+Only its source_url is used to look the work up -- copy that exactly and never
+invent or edit a URL (the title and other fields are taken from the tool data,
+so do not worry about reproducing them perfectly). Finish by calling
+submit_profile exactly once.
 """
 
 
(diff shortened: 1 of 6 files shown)


src/research_ai/services/researcher_profile/agent.py
```diff
@@ -69,41 +71,41 @@ def _user_prompt(expert) -> str:
     )
 
 
-def _ground_works(
-    works: list[dict], toolset: OpenAlexToolset
-) -> tuple[list[dict], list[str]]:
-    """Keep only works whose ``source_url`` the tools actually returned.
+def _ground_works(works, toolset: OpenAlexToolset) -> tuple[list[dict], list[str]]:
```
yattias, line 74: This function seems to filter out a bunch of works that are duplicates or improper. I wonder if we should rename to `_filter_works` or `select_works` or something along those lines because it took me a bit to figure out the function's purpose

yattias: Couple of questions about the `toolset` property

- Can the model invoke the OpenAlex API directly via the toolset?
- Is `_ground_works` used as helper function after results are returned from openalex?

koutst: The filtering is just a side effect of grounding, each kept work is rebuilt from the real OpenAlex record the tools returned, not trusting what the model emitted. A hallucinated work has no matching record, so it falls out of the list automatically. _filter_works/_select_works would imply it returns the model's works unchanged, hiding the fact that the output is reconstructed from Openalex data.

koutst: 1. Can the model invoke OpenAlex directly?
No, the model never touches the OpenAlex API or client. It only calls the tools the toolset exposes (search_institutions, search_authors, get_author, get_author_works), and those handlers are the only code that calls OpenAlex. The model just picks from whatever real records those tools return.

2. Is _ground_works a post-processing helper?
Yes. It runs after the agent finishes, in run_profile_agent. As the tools return works, the toolset caches each full record in returned_works (keyed by source_url). When the model calls submit_profile, we don't trust the works it echoed back. _ground_works re-looks-up each one in returned_works and rebuilds it from the cached ground-truth record, dropping anything without a match. So it's the final grounding step that closes the loop between "model selected" and "tool actually returned."

PR #3704, opened 2026-06-26
Proposal draft driver

koutst:


src/researchhub_document/related_models/researchhub_post_model.py
```diff
@@ -290,7 +291,7 @@ def get_boost_amount(self):
 
     def get_full_markdown(self):
         try:
-            if self.document_type == DISCUSSION:
+            if self.document_type in RESEARCHHUB_POST_DOCUMENT_TYPES:
```
koutst, line 294: We don't actually store the eln_src anymore

PR #3716, opened 2026-06-29
Research Journey Tracker

michaelcanova:
## Overview ##

This PR does several things, first there are some optimizations for the new journal feed to help with the pizza tracker.  Second, this introduces the endpoint for the pizza tracker, which also provides work urls for each stage, so we can easily access data for each stage of the pizza tracker when a segment is clicked, to show the appropriate data, rather than surfacing all of the data for each stage all in a single pull.


src/researchhub_document/serializers/research_journey_tracker_serializer.py
```diff
@@ -0,0 +1,146 @@
...
+                    latest_stage,
+                    visible_post_ids,
+                )
+                for stage in JOURNEY_TRACKER_STAGES
+            ],
+        }
+
+    def get_stage_posts(
+        self, journey: ResearchJourney
+    ) -> dict[str, ResearchhubPost]:
```
koutst: A serializer is supposed to return json, not a Django model

michaelcanova: Good callout, that is my fault even though its really just a small helper function, that it wasn't set as private, and named with the "get" label.  I simplified it down though.  Updated and thank you!

PR #3752, opened 2026-07-05
Journal Feed V2

michaelcanova:
## Overview ##

This is the new journal feed, that pulls completed proposals and registered reports.  It provides the attached grant id, if any, proposal data, review data, and the registered report data to the front end, with sorting options for best, newest, and average peer review score.


src/feed/views/journal_v2_feed_view.py
```diff
@@ -0,0 +1,180 @@
...
+                Q(document_type=REGISTERED_REPORT)
+                | Q(
+                    document_type=PREREGISTRATION,
+                    has_registered_report=False,
+                )
+            )
+        )
+        return queryset
+
+    def build_journal_stage_queryset(self) -> QuerySet:
```
gzurowski: Nit: Private?

michaelcanova: Good callout, I missed that, and it is now fixed.  Thank you again!

PR #3760, opened 2026-07-06
Add logging to loop and drafting

koutst:


gzurowski (approved):
LGTM 👍 

You can also fold some of the data into the `extra` argument of the logging functions. These will become available as structured logging data.

PR #3789, opened 2026-07-13
Registered Report Work Page and Optimizations

michaelcanova:
## Overview ##

This PR creates the APIs required for the work page of a registered report.  It sends back the required data, and also the urls/data for the tracker itself.  We also decided to not show funded/completed proposals in the new journal feed, so that code has been removed.  This also optimizes some of the previous code when creating the note for the registered report.


src/researchhub_document/services/registered_report_work_service.py
```diff
@@ -0,0 +1,91 @@
...
+    grant: ResearchhubPost | None
+    proposal: ResearchhubPost | None
+
+
+class RegisteredReportWorkService:
+    """Load registered report work-page data with viewer-safe tracker stages."""
+
+    def __init__(
+        self,
+        post_model: type[ResearchhubPost] | None = None,
```
gzurowski: The model is not a dependency you want to inject.

gzurowski (changes requested):

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -88,6 +95,32 @@ def create(self, request, *args, **kwargs):
...
         return self.upsert_researchhub_posts(request)
 
+    @action(
+        detail=True,
+        methods=["get"],
+        url_name="registered-report-work",
+        url_path="registered_report_work",
+    )
+    def registered_report_work(
+        self, request: Request, pk: int | None = None
```
gzurowski: `pk` will be of type `str` here.

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -88,6 +95,32 @@ def create(self, request, *args, **kwargs):
     def update(self, request, *args, **kwargs):
         return self.upsert_researchhub_posts(request)
 
+    @action(
+        detail=True,
+        methods=["get"],
+        url_name="registered-report-work",
+        url_path="registered_report_work",
```
gzurowski: These are redundant because DRF will generate the same values. We use them only when the path name differs from the method name. Here, they're just unneeded boilerplate.

src/researchhub_document/services/registered_report_work_service.py
```diff
@@ -0,0 +1,91 @@
...
+        proposal = self.journey_service.get_proposal(journey) if journey else None
+        grant = journey.grant_post if journey else None
+
+        return RegisteredReportWorkPayload(
+            report=report,
+            grant=self.get_visible_stage(grant, user),
+            proposal=self.get_visible_stage(proposal, user),
+        )
+
+    def get_visible_stage(
```
gzurowski: Looks like it should be a private method.
```suggestion
    def _get_visible_stage(
```

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -557,6 +599,11 @@ def update_existing_researchhub_posts(self, request):
             authors = data.get("authors", [])
             rh_post_id = data.get("post_id", None)
             rh_post = ResearchhubPost.objects.get(id=rh_post_id)
+            if rh_post.document_type == REGISTERED_REPORT:
+                return Response(
+                    {"error": "Published registered reports cannot be edited."},
```
gzurowski: `detail` would be a more appropriate message key since this is what DRF's errors serialize to.

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -557,6 +599,11 @@ def update_existing_researchhub_posts(self, request):
...
+                    status=400,
```
gzurowski: An HTTP 400 error indicates a client-side error, such as a malformed request. This isn't the case here. An HTTP 409 might be a better fit.

src/note/views/note_view.py
```diff
@@ -560,6 +560,11 @@ def create(self, request, *args, **kwargs):
         self.kwargs["pk"] = note_id
 
         note = self.get_object()
+        if getattr(note, "post", None) is not None:
+            return Response(
+                {"error": "Published registered report content cannot be edited."},
+                status=400,
```
gzurowski: An HTTP 400 error indicates a client-side error, such as a malformed request. This isn't the case here. An HTTP 409 might be a better fit.

src/note/views/note_view.py
```diff
@@ -560,6 +560,11 @@ def create(self, request, *args, **kwargs):
         self.kwargs["pk"] = note_id
 
         note = self.get_object()
+        if getattr(note, "post", None) is not None:
+            return Response(
+                {"error": "Published registered report content cannot be edited."},
```
gzurowski: `detail` as the message key would be aligned with DRF's error serialization.

src/note/views/note_view.py
```diff
@@ -560,6 +560,11 @@ def create(self, request, *args, **kwargs):
         self.kwargs["pk"] = note_id
 
         note = self.get_object()
+        if getattr(note, "post", None) is not None:
```
gzurowski: This check feels too generic: Does it really apply to all notes? If yes, maybe add a short comment explaining it.

src/researchhub_document/services/registered_report_work_service.py
michaelcanova: Good callout, a misstep on my part, Updated and thank you!

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: Makes sense to me, updated it and thank you!

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: Yea that makes sense, removed and updated.  Thank you!

src/researchhub_document/services/registered_report_work_service.py
michaelcanova: Good callout, I missed that one. Updated and thank you!

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: Ok, I updated it for detail instead of error, thank you again!

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: Understood, set it for a 409 instead of a 400.  Thank you!

src/note/views/note_view.py
michaelcanova: Understood, set it for a 409 instead of a 400.  Thank you!

src/note/views/note_view.py
michaelcanova: Ok, I updated it for detail instead of error, thank you again!

src/note/views/note_view.py
michaelcanova: Yea, I updated it to better target just registered reports, while also fixing the issue from the claude code review above.  Thank you!

src/researchhub_document/services/registered_report_work_service.py
```diff
@@ -22,19 +22,17 @@ class RegisteredReportWorkService:
...
-        post_model: type[ResearchhubPost] | None = None,
         journey_service: JourneyService | None = None,
     ) -> None:
         """Initialize the service with optional dependencies."""
-        self.post_model = post_model or ResearchhubPost
         self.journey_service = journey_service or JourneyService()
 
     def get_work_payload(
-        self, post_id: int, user: User | None
+        self, post_id: str | int, user: User | None
```
gzurowski: A post ID is probably always of type `int`.

gzurowski (approved):

src/researchhub_document/services/registered_report_work_service.py
michaelcanova: Yea, i kept it able to support a str just incase, but that is probably overkill, so I reverted it back to just an int.  Thank you!

PR #3806, opened 2026-07-15
Registered Report Backfill

michaelcanova:
## Overview ##

This PR creates a backfill service for completed proposals, adding a Research Journey to them, allowing them to be turned into a registered report.  Notifications will be sent out as well, but that will be with the next PR, that will include all sent notifications.  Also this PR optimizes some portions of previous PRs for better consistency.


src/researchhub_document/tests/test_backfill_research_journeys.py
```diff
@@ -0,0 +1,349 @@
...
+from researchhub_document.models import (
+    ResearchhubPost,
+    ResearchhubUnifiedDocument,
+    ResearchJourney,
+)
+from researchhub_document.related_models.constants.document_type import PREREGISTRATION
+from user.tests.helpers import create_random_default_user
+
+
+class BackfillResearchJourneysCommandTests(TestCase):
```
gzurowski: For a one-off command, it's probably not worth adding these tests. Since the business logic is in a dedicated service, testing the service should be sufficient.

gzurowski (approved):
If this is a one-off operations, we should have indicator for later cleanup.

src/researchhub_document/tests/test_backfill_research_journeys.py
michaelcanova: Makes sense, and I removed the tests.  Thank you!

PR #3819, opened 2026-07-20
Registered Report Moderation

michaelcanova:
## Overview ##

This PR changes the registered report creation process in two ways:

First, the registered report will now create its own DOI, rather than relying on the propolsal's DOI.

Second the creation process is no longer from the user clicking a notification, but rather created by a moderator/editor, and a new tab/endpoint API was created.

The backfill is also updated to only populate for a few completed proposals for now, and will be updated again later to run all of the rest.


src/feed/tests/views/test_pending_moderation.py
```diff
@@ -172,6 +192,119 @@ def test_counts_grouped_by_type(self):
...
+            "candidate_moderator",
+            moderator=True,
+        )
+        self.author = create_random_default_user("candidate_author")
+        self.client = APIClient()
+        self.client.force_authenticate(user=self.moderator)
+        self.url = reverse("moderator_feed-registered-report-candidates")
+        self.journey_service = JourneyService()
+        RscExchangeRate.objects.create(
+            price_source=MORALIS,
```
koutst: MORALIS is a legacy provider that we will remove, use COIN_GECKO instead

michaelcanova: Switched it over, thank you!

src/feed/views/moderator_feed_view.py
```diff
@@ -105,6 +111,64 @@ def pending_moderation_counts(self, request: Request) -> Response:
...
+        serializer = self.get_serializer(feed_entries, many=True, context=context)
+        return self.get_paginated_response(serializer.data)
+
+    @action(
+        detail=False,
+        methods=["post"],
+        permission_classes=[UserIsEditor | IsModerator],
+        url_path="create_registered_report_draft",
+    )
+    def create_registered_report_draft(self, request: Request) -> Response:
```
koutst: Not sure the actual answer but does this endpoint belong here? Seems like it's creating a note/journal draft? Does that belong in a note or journal admin view or something like that?

src/researchhub_document/views/researchhub_post_views.py
```diff
@@ -327,7 +262,14 @@ def create_researchhub_post(self, request):
         grant_amount = data.get("grant_amount")
         grant_id = data.get("grant_id")
 
-        if authors and request.user.author_profile.id not in authors:
+        if document_type == REGISTERED_REPORT and not request.user.moderator:
```
koutst: Won't this break for editors?

src/feed/views/moderator_feed_view.py
michaelcanova: Good callout, I moved it over to the note!  Thank you!

src/researchhub_document/views/researchhub_post_views.py
michaelcanova: Also a good callout, I fixed it for the editors!  Thank you!

src/note/views/note_view.py
```diff
@@ -185,6 +191,34 @@ def create(self, request, *args, **kwargs):
...
+    @action(
+        detail=False,
+        methods=["post"],
+        permission_classes=[UserIsEditor | IsModerator],
+        url_path="create_registered_report_draft",
+    )
+    def create_registered_report_draft(self, request: Request) -> Response:
+        """Create an editor- or moderator-owned registered report notebook draft."""
+        serializer = RegisteredReportDraftSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
```
koutst: Maybe update RegisteredReportDraftSerializer to RegisteredReportDraftRequestSerializer?

src/note/views/note_view.py
```diff
@@ -185,6 +191,34 @@ def create(self, request, *args, **kwargs):
...
+        except ValueError as error:
+            return Response(
+                {"error": str(error)},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        response_data = NoteSerializer(draft.note, context={"request": request}).data
+        response_data["fundraise_id"] = draft.fundraise.id
+        response_data["journey_id"] = draft.journey.id
+        response_data["proposal_id"] = draft.proposal.id
```
koutst: This belongs in a serializer. Maybe something like RegisteredReportDraftResponseSerializer

koutst: I think this endpoint might belong in the researchhub_document app, since it's more than just a note and that's where the rest of the RR logic is. Something like

```
researchhub_document/
  views/registered_report_draft_view.py
  serializers/registered_report_draft_serializer.py
  services/journal_entry_service.py
  tests/test_registered_report_draft_api.py
```

What do you think?

koutst (approved):

src/note/views/note_view.py
michaelcanova: Sounds good and updated, I also moved the location as mentioned in the next review comment.  Thank you!

src/note/views/note_view.py
michaelcanova: I agree with that, it will look a lot cleaner and more organized.  Updated and thank you again!

PR #3827, opened 2026-07-21
Persist proposal draft agent histories

koutst:
## Stack

1. #3843 — Persist agent execution lifecycles
2. #3844 — Add agent conversation services
3. #3831 — Add agent chat orchestration
4. **#3827 — Persist proposal draft agent histories**

## What changed

- Create and link a durable workflow conversation when proposal drafting begins.
- Start the database-backed execution recorder with a configuration snapshot.
- Expose the Bedrock default model identifier for accurate configuration snapshots.
- Preserve the headless backend prompt without creating artificial chat messages.
- Attach the completed proposal note to the existing workflow conversation.
- Keep persistence failures observational so proposal behavior remains unchanged.

## Why

Proposal drafting is the first production caller of the reusable persistence stack and needs inspectable execution history without coupling the core agent to Django.

## Impact

Completed or failed proposal drafts retain their existing behavior while gaining durable history and notebook association.

## Validation

- Full stacked `research_ai` suite: 651 tests passed.
- Ruff lint and formatting checks passed across `src`.
- Rebase onto current `main` verified with `git range-diff`; all four patches are unchanged.

PR #3830, opened 2026-07-21
Add agent conversation persistence models

koutst:
## Stack

1. **#3830 — Add agent conversation persistence models**
2. #3831 — Add durable agent persistence services
3. #3827 — Persist proposal draft agent histories

This PR is layer 1 of 3 and targets `main`.

## What changed

- Add durable conversation, execution, user-facing message, and execution-trace models.
- Add deterministic per-conversation and per-execution sequence constraints and indexes.
- Add lifecycle, configuration, usage, latency, failure, and terminal-output fields.
- Add a note–conversation join model with real foreign keys and pair uniqueness.
- Keep conversation provenance independent from the notebook where the conversation is exposed.
- Link proposal drafts to agent conversations with `SET_NULL` retention behavior.
- Include focused model coverage and the reversible Django migration.

## Why

Agent activity needs durable domain boundaries that support headless workflows today and resumable notebook conversations, retries, and regeneration later. The join table lets proposal-originated and note-originated conversations use the same notebook lookup without coupling the core conversation model directly to notes.

## Impact

This adds the persistence schema only; it does not yet wire proposal drafting into the database recorder. Deleting either a note or conversation removes only the join row and does not cascade through the association to the other domain object.

## Validation

- Full `research_ai` suite: 639 tests passed.
- Ruff lint and formatting checks passed across `src`.
- Django reported no migration drift.


src/research_ai/tests/test_models.py
```diff
@@ -25,3 +36,108 @@ def test_orcid_handles_plain_string_sources_and_misses(self):
...
+            AgentExecutionMessage.objects.create(
+                conversation=other_conversation,
+                execution=execution,
+                sequence=1,
+                execution_sequence=1,
+                role="user",
+                provenance=AgentExecutionMessage.Provenance.BACKEND,
+                content=[],
+            )
+        self.assertFalse(AgentExecutionMessage.objects.exists())
```
gzurowski, line 60: Except for this one, the other tests are testing Django model mechanics. Is it worth keeping them?

gzurowski (approved):

src/research_ai/tests/test_models.py
koutst: Good catch, will remove

PR #3899, opened 2026-07-30
Bugfix: Increasing image file name length for cover images

michaelcanova:
## Overview ##

This PR updates the maximum image file name length for cover images on notebook publishing.

### DIFF
--- a/src/researchhub_document/tests/test_models.py
+++ b/src/researchhub_document/tests/test_models.py
@@ -230,6 +230,7 @@ class ModelTests(TestCase):
         # Assert
         self.assertEqual(markdown, "Discussion body")
 
+
 class ResearchhubPostStatusTests(TestCase):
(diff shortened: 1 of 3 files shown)


src/researchhub_document/tests/test_models.py
```diff
@@ -230,6 +236,35 @@ def test_get_full_markdown_reads_discussion_body(self):
...
+            for document_type in document_types
+        ]
+
+        # Assert
+        self.assertGreater(len(cover_image_url), 200)
+        self.assertEqual(
+            [post.preview_img for post in posts],
+            [cover_image_url] * len(document_types),
+        )
+
```
gzurowski: I don't think we need to introduce tests verifying basic Django ORM mechanisms.

gzurowski (approved):

PR #3934, opened 2026-08-05
Don't export base email context and merge within sender function

gzurowski:


src/mailing_list/tests/test_lib.py
```diff
@@ -120,6 +120,14 @@ def test_each_recipient_gets_its_own_unsubscribe_url(self):
         first, second = (msg.extra_headers["List-Unsubscribe"] for msg in mail.outbox)
         self.assertNotEqual(first, second)
 
+    def test_renders_assets_base_url_without_the_caller_supplying_it(self):
```
yattias, line 123: I think this test is duplicated

gzurowski: This is not a duplicate. It is located in a different test suite. One is for transactional emails, and the other is for regular emails.

PR #3953, opened 2026-08-06
Reclaim abandoned agent executions and support cancelling a turn (split into #3955 + #3956)

koutst:
last_activity_at was written on every durable write and read by nothing. A worker killed between claiming an execution and recording its terminal status left the row RUNNING, and since a conversation permits one active execution, every later turn on that note was refused as busy -- permanently. A queued attempt the broker never delivered strands a conversation the same way.

A five-minute sweep now reclaims both, marking each INTERRUPTED with its own error type so a stall stays distinguishable afterwards from a run that failed while someone was still driving it. Candidates are picked without a lock and rechecked under one, because the scan races the very workers it hunts: a run that wrote a heartbeat in between is healthy and left alone.

Cancellation reuses that seam and finally sets CANCELLED, the status the context service already understood and nothing ever wrote. It is cooperative -- the worker notices at its next durable write, where the recorder refuses to extend a terminal execution, and unwinds through its ordinary failure path. No error fields are recorded, since stopping a turn is the user's decision rather than a failure, and cancelling when nothing runs is a success reporting cancelled: false, because the client cannot know which side of that race it is on when the user clicks stop.


koutst:
Good catch, and it is worse than described — fixed in 59b1ff2af.

The profile build is not just work that skips the heartbeat, it is a **16-iteration agent run** (`_MAX_ITERATIONS = 16` in `researcher_profile/agent.py`) that writes nothing against the proposal execution, and there are no Celery task time limits bounding it. So a healthy draft really could sit on its creation-time heartbeat past 15 minutes and get failed by its own safety net.

The underlying defect was that reading `last_activity_at` imposes a contract on every execution creator — *a healthy RUNNING row is touched more often than the timeout* — and the sweep assumed that contract rather than establishing it. Filtering to `notebook_chat` would have narrowed the blast radius without fixing that, and would have left `proposal_draft` with the stranding bug this PR exists to kill.

What changed instead:

- **Timeout sized to the longest healthy gap** rather than to one model turn: 15m → 60m, with the reasoning recorded at the constant. Erring long only delays recovery of a genuinely stuck conversation; erring short fails live runs.
- **The contract is now stated** in the module docstring, where the next workflow author will hit it.
- **`DatabaseAgentRecorder.heartbeat()`** for callers doing slow work around the loop. Scoped to non-terminal rows, so it can never revive a cancelled run and returns `False` instead. The proposal runner uses it across the RFP fetch; the profile build itself no boundary heartbeat can shorten, which is what the timeout is for.

Three regression tests, including one at proposal-setup scale (40-minute gap, default timeout) that fails under the old 15-minute value — verified, not assumed.

One thing I did **not** do: `_build_agent` already calls `set_system_prompt`, which stamps the heartbeat, so a second heartbeat immediately after the profile build would be redundant. Only the one covering the RFP fetch is there.

PR #4003, opened 2026-08-17
Keep preprint peer-review bounties in the activity feed

nicktytarenko:
## Summary
- Allow peer-review bounty opportunities on preprints to appear in the activity feed.


nicktytarenko:
not needed

PR #4006, opened 2026-08-17
Notebook Author Sorting

michaelcanova:
## Overview ##

This PR adds a sorting system through a new note database field to allow proper sorting of authors when creating/updating a work in the notebook.

This PR effectively updates the following:
- Authors can be sorted in the notebook itself on the front end through a drag and drop system
- Feeds now properly show authors in the proper order
- Work pages now show authors in the proper order
- Searches will now show authors in the proper order


src/researchhub_document/services/researchhub_post_author_service.py
```diff
@@ -0,0 +1,98 @@
...
+
+def replace_authors(
+    post: ResearchhubPost,
+    authors: list[Author],
+) -> None:
+    """Atomically replace a saved post's authors in the supplied order."""
+    if post.pk is None:
+        raise ValueError("Post must be saved before replacing authors.")
+
+    database = post._state.db or DEFAULT_DB_ALIAS
```
koutst: What is this for? Using Django's private fields is probably not the right approach

michaelcanova: You are correct, there were better implementation methods that I missed.  Thank you again!

koutst:
Take this with a grain of salt but could we make ordering the default model behavior instead of requiring every caller to use list_authors()? For example, expose ordered author_links and derive post.authors from them. What do you think?

PR #4041, opened 2026-08-24
My Funding Page Activity Feed Updates

michaelcanova:
## Overview ##

This PR adds the new `activity_feed/user_activity` endpoint for the activity feed, so we can get a better and  more focused user activity feed for My Funding.


src/feed/views/activity_feed_view.py
```diff
@@ -61,14 +65,18 @@ class ActivityFeedViewSet(FeedViewMixin, ModelViewSet):
...
       - grant_id: all activity on a grant and its applied preregistrations
-      - funder_id: all activity on OPEN/COMPLETED grants where this user is
-        the creator or a contact, plus every preregistration applied to
-        those grants.
       - content_type: RHCOMMENTMODEL, RESEARCHHUBPOST, PAPER, etc.
 
     Filters can be combined: e.g. ?scope=grants&content_type=RHCOMMENTMODEL
     returns only comments across all grant-related documents.
 
+    ``GET user_activity/?user_id=`` returns activity on every public document the
```
yattias: can we put this comment inside the endpoint?

src/feed/views/activity_feed_view.py
```diff
@@ -149,6 +157,32 @@ def list(self, request, *args, **kwargs):
...
+        url_name="user-activity",
+        permission_classes=[IsAuthenticated],
+    )
+    def list_user_activity(self, request: Request) -> Response:
+        """Return activity on every document the user is involved with."""
+        query_serializer = UserActivityQuerySerializer(data=request.query_params)
+        query_serializer.is_valid(raise_exception=True)
+        user_id = query_serializer.validated_data["user_id"]
+
+        if user_id != request.user.id and not request.user.is_moderator_or_editor():
```
yattias, line 174: To separation of concerns clean, let's create a specific permission for this endpoint and do this sort of thing there

yattias (approved):

src/feed/views/activity_feed_view.py
michaelcanova: Moved it into the endpoint, thank you!

src/feed/views/activity_feed_view.py
michaelcanova: I separated it out to its own file within feed, so now we can call that anywhere in the feed if we need it!  Thank you!

PR #4053, opened 2026-08-27
[Notebook Chat] Reap stale agent executions and bound turn tasks

koutst:
## What

First of a 3-PR stack making notebook agent failure a non-event (this → auto-retry → error taxonomy + retry endpoint).

A worker death (OOM, deploy, SIGKILL) leaves its `AgentExecution` RUNNING forever, and the `ra_agent_exec_one_active` constraint then 409-blocks every future message on that conversation permanently. A lost enqueue can strand a PENDING row the same way. There was no liveness sweep (the docstring in `chat_service.py` referenced one that didn't exist), no time limits on the turn task, and no `research_ai` beat entries — two zombie rows have been RUNNING in dev since Jul 31.

## How

- **`AgentExecutionReaperService`** (new, `agent_persistence/reaper_service.py`): finds RUNNING executions whose `last_activity_at` heartbeat is stale (15 min, `updated_date` fallback for null heartbeats) and PENDING rows older than 15 min, and seals them INTERRUPTED with error fields. Staleness is re-checked under `select_for_update(skip_locked=True)`, so a heartbeat that advances between scan and seal — or a row a worker is writing that instant — can never be reaped. Sweeps **all** workflows sharing the models (notebook chat and proposal drafting).
- **Cancel-service reuse**: `AgentExecutionCancelService` grows a shared `_seal()` behind `cancel()` (unchanged) and a new `interrupt()`, so both sealing paths run the same reconciliation — a reaped PENDING turn's prompt is seeded into context, a zombie's unpublished closing answer is dropped.
- **Turn task bounds**: `run_notebook_chat_turn_task` gets `soft_time_limit=900`, `time_limit=960`, and `acks_late=True`. The soft limit lets `run_turn`'s safety net seal the execution FAILED itself; `acks_late` hands a prefetched-but-unclaimed turn to another worker on worker loss (`claim_pending` already makes duplicate delivery a no-op). The reaper threshold is deliberately ≥ the limits, so only truly dead runs can look stale.
- **Beat entries**: reaper every 5 min (worst-case unbrick ~20 min) and `AgentRetentionService.delete_stale_traces` daily — observational trace rows older than 30 days go, in batches; chat messages and context lineage are untouchable.
- `NotebookChatService.notify_turn_reaped` pushes `turn_failed` and clears the transient stream snapshot for reaped turns (harmless no-op group for headless workflows).

No migrations. One accepted trade-off: reaping a proposal-draft execution doesn't touch the `ProposalDraft` row's own status lifecycle (pre-existing gap, unchanged).

## Testing

- New `tests/agent/test_persistence_reaper.py` (11 tests: stale vs fresh RUNNING/PENDING, context reconciliation, cross-workflow sweep, locked staleness re-check, conversation unblocked afterwards, public error rendering).
- New retention sweep test, `run_notebook_chat_turn_task` wrapper tests (previously untested), and reap/prune task tests with AAA markers.
- Full `research_ai` suite passes on this branch alone (1119 tests); `ruff check` and `ruff format` clean.

### DIFF
--- a/src/research_ai/services/agent_persistence/__init__.py
+++ b/src/research_ai/services/agent_persistence/__init__.py
@@ -12,6 +12,7 @@ from .execution_service import (
     AgentExecutionService,
     AgentStaleRetryError,
 )
+from .reaper_service import AgentExecutionReaperService
 from .recorder import DatabaseAgentRecorder
 from .retention_service import AgentRetentionService
 from .run_details_service import (
@@ -25,6 +26,7 @@ __all__ = [
     "AgentConversationBusyError",
     "AgentConversationService",
     "AgentExecutionCancelService",
+    "AgentExecutionReaperService",
     "AgentExecutionService",
     "AgentRetentionService",
     "AgentRunDetails",
--- a/src/research_ai/services/agent_persistence/cancel_service.py
+++ b/src/research_ai/services/agent_persistence/cancel_service.py
@@ -1,11 +1,14 @@
-"""Cooperative cancellation for agent executions.
+"""Cooperative cancellation and interruption for agent executions.
 
 A conversation permits one active execution, so one that never reaches a
 terminal status refuses every later turn as busy -- what happens when a worker
 dies before recording an outcome, or the broker never delivers a queued attempt.
-Cancelling unsticks it, and is deliberately the only way: no timeout is guessed
-at here, because one provider call can retry inside the vendor SDK for over an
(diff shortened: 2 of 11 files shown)


koutst:
Closing unmerged — we're shipping only the error-taxonomy/manual-retry work in #4055, now retargeted to main as a standalone PR. Manual cancel (which works mid-turn, including after a refresh) covers stuck turns for now; the reaper/time-limits/retention work is preserved on this branch if we revisit.

PR #4054, opened 2026-08-27
[Notebook Chat] Auto-retry transient provider failures with backoff

koutst:
## What

Second PR of the failure-recovery stack (**stacked on #4053** — base is set to its branch so only this increment shows; retarget to `main` after #4053 merges).

A single `overloaded_error` costs the user their whole turn — four in six minutes observed on Aug 13 — while `AgentChatService.prepare_retry` sat fully built with zero production callers. This wires it up for transient failures.

## How

- **Provider-layer classification**: `ProviderError` gains a `retryable` flag. The Claude Platform and OpenRouter adapters (the two catalog providers) classify the wrapped SDK failure: connection errors, timeouts, mid-stream transport drops, 408/409/429, and every 5xx (incl. the Messages API's 529 `overloaded_error`) are transient; auth, config, and invalid-request failures are permanent. The recorder stores the verdict in `error_details` at failure time, since the exception dies with the worker.
- **`prepare_retry(pending=True)`**: creates the retry as a claimable PENDING attempt — the same contract `prepare_turn`/`run_turn` already use — with `retry_of`, the same trigger message, and the same context parent.
- **Automatic retry in `run_turn`**: on a retryable terminal FAILED (never a cancellation or external seal — status is re-checked), the turn requeues itself with backoff (5s, then 30s), capped at 2 per trigger by walking `retry_of`. The retry row is created **before** the backoff starts, so the conversation's busy check spans the whole window: a user message can't interleave with an automatic retry, and losing the creation race to a fresh user message just abandons the retry. Each retry announces itself with `turn_queued` so clients render it as a new attempt.

No migrations; `error_details` is an existing JSON field.

## Testing

- Provider classification tests for both adapters (transient vs permanent, and that `ProviderError.retryable` survives the wrap).
- `PendingRetryPreparationTests`: claimability, lineage reuse, busy-check held, regenerate refused.
- `NotebookChatAutoRetryTests`: backoff schedule and cap, permanent failures and crashes not retried, cancelled turns never overridden, lost races abandoned quietly, busy check across the backoff window, event ordering.
- Full `research_ai` suite passes on this branch (1137 tests); ruff clean.

### DIFF
--- a/src/research_ai/services/agent/errors.py
+++ b/src/research_ai/services/agent/errors.py
@@ -40,7 +40,22 @@ class ProviderError(AgentRunError):
 
     Providers raise this without a transcript; the loop attaches ``messages``
     and ``iterations`` as the error propagates through it.
+
+    ``retryable`` marks transient conditions (overload, rate limit, 5xx, a
+    dropped stream) where repeating the identical request may succeed, as
+    opposed to failures of the request itself (auth, config, invalid input).
     """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        messages: list[Message] | None = None,
+        iterations: int | None = None,
+        retryable: bool = False,
+    ):
+        super().__init__(message, messages=messages, iterations=iterations)
+        self.retryable = retryable
 
 
 class IncompleteTurnError(AgentRunError):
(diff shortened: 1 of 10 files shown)


koutst:
Closing unmerged — scope cut to #4055, which now carries the provider retryable classification directly. The automatic backoff retry is dropped in favor of the readable error + manual Retry affordance.

PR #4057, opened 2026-08-29
Adding new columns to Note table

michaelcanova:
## Overview ##

This PR creates the migration and model updates to add 3 new columns to the note table.

### DIFF
--- a/src/note/related_models/note_model.py
+++ b/src/note/related_models/note_model.py
@@ -22,7 +22,6 @@ class Note(DefaultModel):
     )
     image = models.TextField(
         blank=True,
-        null=True,
     )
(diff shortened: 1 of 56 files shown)


koutst (approved):

src/note/related_models/note_model.py
```diff
@@ -20,12 +20,21 @@ class Note(DefaultModel):
         null=True,
         blank=True,
     )
+    image = models.TextField(
+        blank=True,
+        null=True,
```
koutst: Probably worth addressing, so we just have empty strings instead of having to check for nulls and empty strings

PR #4058, opened 2026-08-29
Adding 3 new Tables for Note

michaelcanova:
## Overview ##

This PR creates the migrations and models adding 3 new tables for saving note details.


koutst (approved):
Remove null=True from text fields to avoid having to check for null and empty string

PR #4088, opened 2026-09-07
Add FundingPool contributions and auto-create on grant create

nicktytarenko:
## Summary
- Add `FundingPoolService.create_contribution` and `FundingPoolViewSet` with `GET /api/funding_pool/{id}/` and `POST .../create_contribution/`
- Auto-create a `FundingPool` when a Grant is created
- Support optional `funding_pool_id` on payment intents (mutually exclusive with `fundraise_id`); on Stripe success, contribute to the pool via funding credits

### DIFF
--- a/src/purchase/services/funding_pool_service.py
+++ b/src/purchase/services/funding_pool_service.py
@@ -1,0 +1,178 @@
+import logging
+from decimal import Decimal
+
+from django.contrib.contenttypes.models import ContentType
+from django.db import transaction
+
+from analytics.tasks import track_revenue_event
+from purchase.models import Balance, FundingPool, Grant, Purchase, RscExchangeRate
+from purchase.related_models.constants import (
+    MAXIMUM_FUNDRAISE_CONTRIBUTION_AMOUNT_RSC,
+    MINIMUM_FUNDRAISE_CONTRIBUTION_AMOUNT_RSC,
+)
+from purchase.related_models.constants.currency import RSC
+from purchase.services.fundraise_service import FundraiseService
+from reputation.models import BountyFee
+from reputation.utils import calculate_bounty_fees, deduct_bounty_fees
+from user.models import User
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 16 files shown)


src/purchase/services/funding_pool_service.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        return pool
+
+    def validate_pool_for_contribution(
+        self, pool: FundingPool
+    ) -> tuple[bool, str | None]:
+        """Validate that a funding pool can accept contributions."""
+        if pool.status != FundingPool.OPEN:
+            return False, "Funding pool is not open"
+        return True, None
```
gzurowski, line 39: Use exceptions to signal errors in an idiomatic way. The Go-style of returning errors is problematic in Python, especially when it's not well documented and relies on plain strings.

Additionally, this could be a simple property on the funding pool model itself, so a client could ask `funding_pool.is_valid_for_contribution`.

src/purchase/services/funding_pool_service.py
```diff
@@ -0,0 +1,178 @@
...
+        if any(
+            not value.is_finite() or value < 0
+            for value in (fee, rh_fee, dao_fee, total_cost)
+        ):
+            return None, "Invalid fee configuration"
+
+        with transaction.atomic():
+            pool = FundingPool.objects.select_for_update().get(id=pool.id)
+            if pool.status != FundingPool.OPEN:
+                return None, "Funding pool is not open"
```
gzurowski, line 109: Not using exceptions here could be even more problematic in the future if anything gets written in the transaction before returning with an error string. In that case, the data would be written silently. Please refactor to use an exception, which is the idiomatic way to handle errors in Python.

nicktytarenko: this is fixed in next PR here
https://github.com/ResearchHub/researchhub-backend/pull/4097/changes/c708f5366837811abf8f8c47694cd0f58c242bf7

